PR #22, opened 2012-06-02
Added support for Arduino Leonardo

Lauszus:
The Teensy 2.0 is no longer the default, as both the Leonardo and the Teensy 2.0 is using the ATmega32u4.
To use the Teensy 2.0 you have to uncomment "#define BOARD_TEENSY" in avrpins.h.

I set the Arduino Leonardo as the default, as I thought most users would use that compared to the Teensy 2.0.


krushnaR:
My PC Does not recognize the Leonardo at all when USB Host Shield is connected to it.

Lauszus:
@krushnaR make sure you haven't shorted out the 5V line.

Issue #95, opened 2014-10-04
Right Shift Key Does not produce Capital Character in Keyboard example Left OK


joyceck:
Right Shift Key Does not produce Capital Character in Keyboard example Left shift OK. Using it on a circuits@home USB mini adapter. and 2.0 lib

Lauszus:
This was fixed in: https://github.com/felis/USB_Host_Shield_2.0/pull/98.

PR #98, opened 2014-10-23
Make the right shift key work in OemToAscii (only the left shift key was...

davidbkemp:
This might be a fix for issue #95.


Lauszus:
Thank you. Sorry for not getting back to you before.

PR #99, opened 2014-10-23
Avoid trademark issues with library name

ivankravets:
Added authors, frameworks and platforms fields

Updated Library Registry: http://platformio.ikravets.com/#!/lib/show/59/USB-Host-Shield-20

### DIFF
--- a/library.json
+++ b/library.json
@@ -1,7 +1,29 @@
 {
-  "name": "Arduino-USBHostShield20",
+  "name": "USB-Host-Shield-20",
   "keywords": "usb, hid",
-  "description": "Revision 2.0 of MAX3421E-based USB Host Shield Library for AVR's",
+  "description": "Revision 2.0 of MAX3421E-based USB Host Shield Library",
+  "authors":
+  [
+    {
+      "name": "Oleg Mazurov",
+      "email": "mazurov@circuitsathome.com",
+      "url": "http://www.circuitsathome.com",
+      "maintainer": true
+    },
+    {
+      "name": "Alexei Glushchenko",
+      "email": "alex-gl@mail.ru"
+    },
+    {
+      "name": "Kristian Lauszus",
+      "email": "kristianl@tkjelectronics.com",
+      "url": "http://tkjelectronics.com"
+    },
+    {
+      "name": "Andrew Kroll",
+      "email": "xxxajk@gmail.com"
+    }       
+  ],
   "repository":
   {
     "type": "git",
@@ -11,5 +33,7 @@
   [
   	"examples/*/*.ino",
   	"examples/*/*/*.ino"
-  ]
+  ],
+  "frameworks": "arduino",
+  "platforms": "atmelavr"
 }


ivankravets:
@Lauszus Sorry for the late answer, but I have great news - PlatformIO 1.0 (production release) has been released! [CHANGES](http://docs.platformio.org/en/latest/history.html). 

It received support for the new platforms and boards, including [Atmel SAM](http://platformio.org/#!/platforms/atmelsam) and [Teensy](http://platformio.org/#!/platforms/teensy).

Do you have Arduino Due? :blush:

xxxajk:
I have recently acquired a Due.

On Fri, Feb 27, 2015 at 5:32 PM, Ivan Kravets <notifications@github.com>
wrote:

> @Lauszus <https://github.com/Lauszus> Sorry for the late answer, but I
> have great news - PlatformIO 1.0 (production release) has been released!
> CHANGES <http://docs.platformio.org/en/latest/history.html>.
>
> It received support for the new platforms and boards, including Atmel SAM
> <http://platformio.org/#!/platforms/atmelsam> and Teensy
> <http://platformio.org/#!/platforms/teensy>.
>
> Do you have Arduino Due? [image: :blush:]
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/felis/USB_Host_Shield_2.0/pull/99#issuecomment-76485258>
> .
>



-- 
Visit my github for awesome Arduino code @ https://github.com/xxxajk

ivankravets:
@xxxajk Could I ask you to test it with new PlatformIO? I'm interested in "uploading". I didn't test it :(

Thanks in advance! :+1:

xxxajk:
I'll place it into my to-do queue.

On Sat, Feb 28, 2015 at 3:06 PM, Ivan Kravets <notifications@github.com>
wrote:

> @xxxajk <https://github.com/xxxajk> Could I ask you to test it with new
> PlatformIO? I'm interested in "uploading". I didn't test it :(
>
> Thanks in advance! [image: :+1:]
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/felis/USB_Host_Shield_2.0/pull/99#issuecomment-76543388>
> .
>



-- 
Visit my github for awesome Arduino code @ https://github.com/xxxajk

xxxajk:
You need to parse the includes better.
The top of the source reads:
```
#include <usbhub.h>
// Satisfy IDE, which only needs to see the include statment in the ino.
#ifdef dobogusinclude
#include <spi4teensy3.h>
#endif
```
and your generated cpp file incorrectly did this:
```
#include <Arduino.h>
#include <usbhub.h>
#include <spi4teensy3.h>
```

The idea is to hint to the IDE that it *might* need this header.
If the IDE *finds* the header in the paths, *then* it just includes the path into the include path search.
The Arduino IDE will look for it, but not find it, and that is fine and happy, as it doesn't need it on the due, or avrs, etc.

xxxajk:
Other hints, if I may...
There is no need to re-write the files. Here is why...
1: GCC has an option that can do this include from the command line, so I don't understand why you are doing it that way. Furthermore, it is bad practice to not have Arduino.h included in your sketch, I force people to do the include in their code.
2: GCC can totally accept the .ino extension. Just tell GCC the language.

xxxajk:
One last hint for the future. Don't re-arrange the code, and don't prototype stuff. It isn't needed. If you do that, it will totally break the next version of the library because what is included depends on defines at the *top* of the file *before* and include statements.

ivankravets:
> GCC has an option that can do this include from the command line, so I don't understand why you are doing it that way. Furthermore, it is bad practice to not have Arduino.h included in your sketch, I force people to do the include in their code.

I don't use *.ino or *.pde formats, I use *.c/.*cpp. However, PlatformIO should support all formats. I can't say each time that "someone" should include `Arduino.h` in `*.ino` sketch. I don't know who has "invented" this format, but it was bad idea :(

> 2: GCC can totally accept the .ino extension. Just tell GCC the language.

I know it. Nevertheless, the most of *.ino files are invalid by default. They are not suitable C/CPP source files. As you mentioned, some of files don't contain `Arduino.h`, others don't have defined prototypes. The problem is more complicated then "change *.ino extension to *.cpp". PlatformIO does it "on the fly".

On the whole, I will rewrite the part of code which converts INO to CPP. It will have more clear behaviour.
I've just opened new issue https://github.com/ivankravets/platformio/issues/100 for it.

Thanks a lot for your time and valuable feedback.

ivankravets:
Did you upload the firmware to your Due via `platformio run --target upload`?

xxxajk:
You want me to re-try with blink?

ivankravets:
If it isn't difficult for your :flushed: You can also try "USB_Host_Shield_2.0" library, but don't use "*ino" file if you have any problems which you described above. Just use plain *.c/cpp.

Thanks in advance!

xxxajk:
That's the trouble. You can't when the includes get re-arranged. 
I'll try blink in a little while, as I just got done tweaking some USB Host code, and it needs to be tested.

ivankravets:
> That's the trouble. You can't when the includes get re-arranged. 

I don't re-arrange *.c/cpp files. Just *.ino/pde

xxxajk:
Don't re-arrange the includes. That's going to cause mega issues.

On Mon, Mar 2, 2015 at 7:32 AM, Ivan Kravets <notifications@github.com>
wrote:

> That's the trouble. You can't when the includes get re-arranged.
>
> I don't re-arrange *.c/cpp files. Just *.ino/pde
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/felis/USB_Host_Shield_2.0/pull/99#issuecomment-76704169>
> .
>



-- 
Visit my github for awesome Arduino code @ https://github.com/xxxajk

ivankravets:
> Don't re-arrange the includes. That's going to cause mega issues.

Which are types of includes do you mean? Which are inside source files or which are passed to C Pre-processor?

xxxajk:
The ones inside ino/pde.

To give you an example of why this is important, here is a small bit of code from the up-comming USB-host code that is not released yet.

```
// inline library loading
#define MAX_USB_MS_DRIVERS 1 // must be 1 to 4
#define LOAD_USB_HOST_SYSTEM
#define LOAD_USB_HOST_SHIELD
#define LOAD_UHS_BULK_STORAGE
#define LOAD_RTCLIB
#define LOAD_GENERIC_STORAGE
#define _USE_MAX3421E_HOST 1
#define _USE_FASTSEEK 0
#define _USE_LFN 3
#define _CODE_PAGE 437
#define	_FS_LOCK 1

#include <Arduino.h>
#ifdef true
#undef true
#endif
#ifdef false
#undef false
#endif

#include <stdio.h>
#include <Wire.h>
#include <SPI.h>
// This figures out how much of the demo we can use
#ifdef __AVR__
#include <avr/io.h>
#endif
```
There's two already.
Later on in the demo code, you have the following:
```
#ifndef USB_HOST_SHIELD_USE_ISR
#if defined(__arm__)
#include <dyn_SWI.h>
#define USB_HOST_SHIELD_USE_ISR 1
#endif
#endif

#include <RTClib.h>
#include <UHS_host.h>
#include <USB_HOST_SHIELD.h>
#include <UHS_HUB.h>
#include <UHS_BULK_STORAGE.h>
#include <UHS_FS.h>
```
Looking at the above bits...
1: If it's arm.... well, I think you see that. Including this on an AVR could cause all sorts of havoc. The reverse is true as well, you don't want to include io/avr.h if the proper ifdef is there either.
2: C++ code is actually loaded in those headers. The settings MUST come before it in the define statements at the top of the sketch. Furthermore they need to be loaded in that specific order.
3: If you attach ```#include <Arduino.h>``` at the beginning as you do now, that's fine. No harm done.
4: If you prototype any methods or C functions that are ifdef'd out, you will run into the same stupid issue the IDE has... namely it pukes out a warning that shouldn't be there. The Arduino team is actually working on fixing this stupid bug.

What you may find more useful is using gcc -E, and take the output from that, instead of attempting to reinvent the wheel.

ivankravets:
I got you, thanks. I mean that you can reproduce these problems only via *.ino format. Why you don't use native *.cpp? Is it related to Arduino IDE and compatibility? Do you have any propositions how I should handle "ino" files, which are not clear "cpp" (without ``Arduino.h``, prototypes and etc)? 

See current implementation: https://github.com/ivankravets/platformio/blob/develop/platformio/builder/tools/platformio.py#L258

I'm agree with you that I have to keep ``#includes`` without any changes, but how about prototypes?

xxxajk:
I would guess that prototypes are possibly fine if you actually follow the #if statements.

ivankravets:
I will not follow/handle any pre-processor's statements, I don't want to be in pre-processor role. I will put each declaration of prototype "_BEFORE_" real code which declares that prototype.

For example `.ino` file:
```cpp
/*... rest of code...  */
#ifdef MYDEFINE
void someProtoype(int arg) {
}
#endif

/* ... rest of code...*/
```

PlatformIO will convert it to `*.cpp`:
```cpp
/*... rest of code...  */
#ifdef MYDEFINE
void someProtoype(int arg);
void someProtoype(int arg) {
}
#endif

/* ... rest of code...*/
```

To my mind, it should fix all issues.

xxxajk:
It might not, for example if someone calls a function or method before it is declared. Not sure how you would handle that any other way.

ivankravets:
Dear @xxxajk,

Thanks a lot for your request. I fixed it in https://github.com/ivankravets/platformio/releases/tag/v1.2.0 release and tested with Arduino Due. Works nice.

xxxajk:
Super :-)

On Sat, Mar 21, 2015 at 7:13 AM, Ivan Kravets <notifications@github.com>
wrote:

> Dear @xxxajk <https://github.com/xxxajk>,
>
> Thanks a lot for your request. I fixed it in
> https://github.com/ivankravets/platformio/releases/tag/v1.2.0 release and
> tested with Arduino Due. Works nice.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/felis/USB_Host_Shield_2.0/pull/99#issuecomment-84313645>
> .
>



-- 
Visit my github for awesome Arduino code @ https://github.com/xxxajk

Issue #109, opened 2014-11-19
PL2303::SndData() not working on PL2303HX chips


v173k:
Hello,

I am using a NS-RAW GPS Module (http://navspark.mybigcommerce.com/ns-raw-carrier-phase-raw-measurement-output-gps-receiver/) with a USB Host Shield. It has a PL2303HX USB to Serial chip (see: https://cloud.githubusercontent.com/assets/277716/5099336/8ccb9090-6f59-11e4-9ada-cf11f11252c2.png )

> USB_Desc Output:
> Device descriptor: 
> Descriptor Length: 12
> Descriptor type:   01
> USB version:       0110
> Device class:      00
> Device Subclass:   00
> Device Protocol:   00
> Max.packet size:   40
> Vendor  ID:        067B
> Product ID:        2303
> Revision ID:       0400
> Mfg.string index:  01
> Prod.string index: 02
> Serial number index:   00
> Number of conf.:   01

It appears that sending data using `PL2303::SndData()` does not actually work even though no error seems to be reported. However, if I use and older version of the library (specifically commit: 3d068a063795505a6b577e31e2955577505f3d4e) and uncomment `#define PL2303_COMPAT` from `cdcprolific.h`  everything works ok.

Oleg has also discussed this here: http://www.circuitsathome.com/mcu/major-acmprolific-bug-fix-posted-on-github but said that it didn't make any difference. I would like to confirm that it indeed does make a difference but only when Sending Data. The current version of the library's `RcvData()` function works fine for receiving data from the Pl2303HX.

I will make a Pull request with my changes (though most of this is Oleg's old code inserted into the latest commit) for review. I just wanted to have this issue recorded for future reference/documentation.

Thanks
Vivek

Lauszus:
This issue was fixed in: https://github.com/felis/USB_Host_Shield_2.0/pull/110.

PR #110, opened 2014-11-19
Fix for PL2303HX chips so that PL2303::SndData() works.

v173k:
This is the fix for the Issue #109 I posted.

This used to work earlier by uncommenting "#define PL2303_COMPAT" from cdcprolific.h

I just re-added code from one of Oleg Mazurov's old commits (see commit: 3d068a063795505a6b577e31e2955577505f3d4e ) which for some reason was lost during a later merge/pull request (See commit: 3d068a063795505a6b577e31e2955577505f3d4e)


Lauszus:
@felis or @xxxajk can you take a look at this at some point, so we can get potentially bugfixes into 3.0 as well.

Lauszus:
Thanks @v173k. It is now merged into master.

PR #135, opened 2015-02-14
Xbox ONE Controller Support

guruthree:
Works with my controller on a off brand USB Host Shield and Leonardo.


XBOXONE.h
```diff
@@ -0,0 +1,180 @@
...
+
+// PID and VID of the different devices
+#define XBOX_VID                                0x045E // Microsoft Corporation
+#define XBOX_ONE_PID                            0x02D1 // Microsoft One Wired controller
+
+#define XBOX_REPORT_BUFFER_SIZE 14 // Size of the input report buffer
+
+// Used in control endpoint header for HID Commands
+#define bmREQ_HID_OUT USB_SETUP_HOST_TO_DEVICE|USB_SETUP_TYPE_CLASS|USB_SETUP_RECIPIENT_INTERFACE
+#define HID_REQUEST_SET_REPORT      0x09
```
Lauszus: These two lines can be removed.

examples/Xbox/XBOXONE/XBOXONE.ino
```diff
@@ -0,0 +1,110 @@
...
+  if (Usb.Init() == -1) {
+    Serial.print(F("\r\nOSC did not start"));
+    while (1); //halt
+  }
+  Serial.print(F("\r\nXBOX USB Library Started"));
+}
+void loop() {
+  Usb.Task();
+  if (Xbox.XboxOneConnected) {
+
```
Lauszus: Could you removed this space?

XBOXONE.cpp
```diff
@@ -0,0 +1,359 @@
...
+                D_PrintHex<uint8_t > (readBuf[i], 0x80);
+                Notify(PSTR(" "), 0x80);
+        }
+        Notify(PSTR("\r\n"), 0x80);
+#endif
+}
+
+uint8_t XBOXONE::getButtonPress(ButtonEnum b) {
+        return (bool)(ButtonState & ((uint16_t)pgm_read_word(&XBOX_BUTTONS[(uint8_t)b])));
+}
```
Lauszus: Would it be possible for you to change this to something similar to the XBOXRECV code: https://github.com/felis/USB_Host_Shield_2.0/blob/master/XBOXRECV.cpp#L409-L415? Then we don't need the LeftTrigger/RightTrigger enums.

XBOXONE.cpp
```diff
@@ -0,0 +1,359 @@
...
+
+bool XBOXONE::getButtonClick(ButtonEnum b) {
```
Lauszus: Could you check for L2/R2 here, as it is done in the XBOXRECV code: https://github.com/felis/USB_Host_Shield_2.0/blob/master/XBOXRECV.cpp#L418-L430. I know it looks ugly, but it should be okay for now. I am thinking of rewriting a lot of the code at some point anyway, then I will fix stuff like this as well.

Lauszus:
Thank you very much!

I have made some inlined comments.

Also thank you for adding documentation as well ;)

XBOXONE.h
```diff
@@ -0,0 +1,180 @@
...
+
+#define XBOX_MAX_ENDPOINTS   3
+
+/** This class implements support for a Xbox wired controller via USB. */
```
Lauszus: Can you update this comment?

guruthree:
I found it very confusing that something that's an analogue input was being treated in the same function as digital input. I didn't even recognise what was happening with L2/R2 until I'd pretty much removed all of that code. Looking now at putting it back in, the triggers on the XBoxOne controller are 0-1023, and so won't fit in an 8-bit return. For compatibility with the other controller libraries I can add the triggers back into the state and press checks, with the latter returning the sensor value / 4. Still, I think getTrigger() is necessary for the extra resolution, and as the trigger really isn't a button.

Lauszus:
This was for historical purposes, as the PS3 controller has many more analog buttons, so instead of adding a getAnalogButton for the Xbox classes only to read the triggers. I simply decided to just use the getButtonPress function. I think it's too late to redo it now and I would like the Xbox classes to be as consistent as possible.

I would be okay if you just changed getButtonPress to return a ```uint16_t```, so we still get the full resolution. I should just be mentioned in the comments, then it will also be included in the documentation as well (http://felis.github.io/USB_Host_Shield_2.0/).

Lauszus:
@guruthree I just merged your pull request. I also added some bug fixes: https://github.com/felis/USB_Host_Shield_2.0/commit/fcab83dcb398ec904db22009d21fa2c1fa133ba1. Please test the newest code on the master branch and confirm that it is working.

Thank you again for your pull request!

PR #137, opened 2015-03-03
Implement Arduino SPI transaction API

Lauszus:
Needs testing...


Lauszus:
@xxxajk can you test this on your Due at some point?

PR #145, opened 2015-04-15
Update Usb.cpp

feilipu:
Typo in the InTransfer function relating to resolving toggle errors arising from hrTOGERR.


felis:
thx!

xxxajk:
no need to test. it is rare that this code would get hit

PR #151, opened 2015-05-17
Corrected spelling mistake "Purble" to "Purple"

Michi302:
I noticed that "Purble" is spelled wrong in multiple files and fixed it. I kept the wrong name to not break projects using it.


Lauszus:
Thanks!

PR #156, opened 2015-06-01
Fixed Compile Error with RedBearLabs nRF51288

sieren:
Fixed a macro definition which would cause the compiler to try
to compile against an undefined function in the RBL SDK.


Lauszus:
Thanks!

PR #157, opened 2015-06-10
Added support for another Mad Catz controller

Lauszus:
See: http://blog.tkjelectronics.dk/2012/07/xbox-360-controller-support-added-to-the-usb-host-library/comment-page-2/#comment-548083


felis:
thx!

Lauszus:
I will merge this once the user has confirmed if it is working or not :)

PR #170, opened 2015-07-21
Add Teensy LC support. Note: disables spi4teensy lib. (Barry Carter)

ginge:
This patch adds support for the Teensy LC.
The spi4teensy lib is not compatibe, so for now this forces the SPI into default SPI lib.
Tested working on a USB host mini 2.0 and a USB host 5v modded to 3v3

Tangential note:
The Teensy LC has lower drive current available on the pins compared with the Teensy 3.x.
This meant that I had to remove the level shifters from the host board completely unlike the 3.x which worked just fine.

### DIFF
--- a/avrpins.h
+++ b/avrpins.h
@@ -835,6 +835,70 @@ MAKE_PIN(P33, CORE_PIN33_PORTREG, CORE_PIN33_BIT, CORE_PIN33_CONFIG);
 
 #undef MAKE_PIN
 
+#elif defined(CORE_TEENSY) && (defined(__MKL26Z64__))
+
+// we could get lower level by making these macros work properly.
+// for now just use the semi optimised version, it costs a lookup in the pin pgm table per op
+// but for now it will do.
+//#define GPIO_BITBAND_ADDR(reg, bit) (((volatile uint8_t *)&(reg) + ((bit) >> 3)))
+//#define GPIO_BITBAND_MASK(reg, bit) (1<<((bit) & 7))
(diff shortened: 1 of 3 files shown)


xxxajk:
I am aware of the LC and level shifter problem as well. It isn't so much a drive issue as it is a level issue. If you power the board with 3.3v instead of 3v, it should work without any modifications too.

xxxajk:
I mean, 3.3v instead of 5v

ginge:
Interesting.
Yeah I had already modded the board to pure 3.3 but it was so marginal that
it would fail every few hundred transactions.
No issues since the level shifter went.

xxxajk:
Another point is that the spi4teesny3 is for teensy3.x, not the LC. I have not had the chance yet to port it. In either case, next UHS lib won't require the spi speedup, but you should try to power it by using 3.3v instead of 5. I would know, since I had an LC before it was publicly released. LC was just never set as full support in 2.0 since 3.0 is in development.

xxxajk:
Ok, this is good to know. It isn't a fact of it being 3.3v or 5v, the problem is the level shifters are speced at 5v on the output side, and simply get too high of a latency.

xxxajk:
Have you tried the ush 3.0 library pre-release using the native usb on the LC yet?

ginge:
no, not yet. I have quite a few hacks into the usb host codebase to support some USB packet forwarding to another device. This will hold porting back a little.
I am starting a new project next week that I will use it with to get familiar with. Cheers.

xxxajk:
One other note... the UHS mini version does not contain level shifting, but you do need to place 5v on the vbus line. It is also more suited for the teensy ARM series because of this fact.

Lauszus:
Thanks!

PR #174, opened 2015-09-10
Fixed toggle issue

ethiebaut:
I had to fix this in order to be able to make the acm_terminal sample working when connecting to a WaveId PcProx Plus. It seems that the toggle wasn't toggled correctly.

Note that the above comment ("yes, we flip it wrong here so that next time it is actually correct!") may not be accurate now.

I also suspect that line 173 is incorrect as it mixes bmRcvToggle and bmSNDTOGRD but I haven't changed it as it seems to work like this for me, so it's probably expected?

### DIFF
--- a/Usb.cpp
+++ b/Usb.cpp
@@ -234,7 +234,7 @@ uint8_t USB::InTransfer(EpInfo *pep, uint16_t nak_limit, uint16_t *nbytesptr, ui
                 rcode = dispatchPkt(tokIN, pep->epAddr, nak_limit); //IN packet to EP-'endpoint'. Function takes care of NAKS.
                 if(rcode == hrTOGERR) {
                         // yes, we flip it wrong here so that next time it is actually correct!
-                        pep->bmRcvToggle = (regRd(rHRSL) & bmRCVTOGRD) ? 0 : 1;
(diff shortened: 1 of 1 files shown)


Lauszus:
@xxxajk can you take a look at this?

blxd:
This fixes an issue I have been having too.

bbx10:
This patch fixes the problem I was having with PS3USB. Previously no buttons and neither joystick worked. With the patch, all buttons and both joysticks are working. 

Thanks!

xxxajk:
Will look when I get a chance.... been busy with UHS30 and work, and chipkit support

baghdasaryan:
This patch helped address a problem with no characters being printed out to console, when using PcProx USB Badge Reader. Further details on the issue I was having and the resolution are available here: https://github.com/felis/USB_Host_Shield_2.0/issues/315

Unless this change has known backwards compatibility issues, I would like to suggest merging it in, as it will solve issues for several projects and prevent new users from spending a bunch of time on debugging.

Otherwise, I think we should update library's [FAQ section](https://github.com/felis/USB_Host_Shield_2.0#faq) to link to this issue (I am ready to do that and send a pull request).

Lauszus:
One that this would be somewhat backwards compatible was to simply use the original code at first and if that fail use this code.

PR #195, opened 2015-12-29
Add possibility to set a custom FTDI PID

romerod:
As there are a lot of FTDI chips arround with a customized PID, i added the possibility to change the PID to be able to connect to customized FTDI chips.


Lauszus:
Thanks :)

PR #207, opened 2016-01-28
Have PS3.printStatusString() print to a new line.

matt-bull:
Have PS3.printStatusString() print to a new line each time it is called, making serial output look much nicer if called multiple times.

### DIFF
--- a/PS3BT.cpp
+++ b/PS3BT.cpp
@@ -150,7 +150,7 @@ String PS3BT::getTemperature() {
 
                 return output;
         } else
-                return "Error";
+                return "\r\nError";
 }
 
 bool PS3BT::getStatus(StatusEnum c) {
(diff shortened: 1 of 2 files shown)


Lauszus:
Can you add this for all the cases? I mean both for the Move controller and in case it prints the error message: https://github.com/matt-bull/USB_Host_Shield_2.0/blob/patch-1/PS3BT.cpp#L199.
Also can you add a carriage return as well i.e. use both ```\r\n```.

If I remember correctly you will have to increase the buffer size by two as well.

matt-bull:
Hows this?

matt-bull:
@Lauszus So, I'm not entirely sure what buffer you're referring to. Do you mean increase the size of the character string by 2 or is there an actual buffer value elsewhere?

Sorry if it's a silly question, my current programing experience is limited to a single college level C course.

PS3BT.cpp
```diff
@@ -150,7 +150,7 @@ String PS3BT::getTemperature() {
 
                 return output;
         } else
-                return "Error";
+                return "\r\nError";
```
Lauszus: Please revert this change.

Lauszus:
@matt-bull yes that is the right buffer. The reason why it has to be 102 is because the maximum length of the message is 101 and then we need one more to store the character.

This simple Python script shows what I mean:
```python
String = '\r\nConnectionStatus: Unplugged - PowerRating: Not Charging - WirelessStatus: Bluetooth - Rumble is off'
print len(String) + 1
```

Please also see my inline-comment.

Lauszus:
I just merged your PR manually: https://github.com/felis/USB_Host_Shield_2.0/commit/5f460dec87e1427fbad110e35b72e1c1d68a4293.

Thanks!

matt-bull:
Oops, please ignore the deletion and restoration of my branch, still getting the hang of github..

Lauszus:
@matt-bull no problem. It's already merged :)

Issue #189, opened 2015-12-06
Missing wheel mouse reports in HID full mode (however HID boot mode reports OK)


joseangeljimenez:
We have been working with several different USB mouse devices, about 10 in total, testing the behaviour of the wheel and how the mouse reports the wheel (Z axis) movements.

What we found so far:
- Even tough the HID boot protocol should not include the wheel byte in the reports, we have found that about 40-50% of our tested devices DO report a fourth byte with the wheel byte counter, which is nice.
- Again, in HID protocol, thouse devices reporting the fourth/wheel byte, do it as expected, without missing any movements coming from the wheel. The reports will include slow and fast wheel movements, without losing events.
- When using the full/complete HID mode, all 100% tested devices report the fourth/wheel byte. However, now the devices will not report correctly the wheel movements. Some of the movements coming from the wheel are missed, not reported. The rest of X, Y, Z movements, however, are timely reported without trouble.
- We have tested the both modes above with the examples included in the USH Host Shield library, with the debug mode active, without adding/modifying a single line of code.

Questions:
- Could this be a subtle bug in the library?
- How can we dig deeper into the subject. Any help is welcome.
- Is there any additional (programatic) step that we have to carry to instruct the mouse to report the wheel in HID full mode? Sending an specific low-level command for the USB device so that it better reports the wheel movement?

Lauszus:
Have you looked at more than the first four bytes? Maybe it is sending out some more information.

Also have you replaced the following line: https://github.com/felis/USB_Host_Shield_2.0/blob/ad966a5fb1bda601daa394da7b8c99af3efa3278/hidboot.h#L459 with:
```C++
rcode = SetProtocol(i, HID_RPT_PROTOCOL);
```

Or are you calling SetProtocol: https://github.com/felis/USB_Host_Shield_2.0/blob/ad966a5fb1bda601daa394da7b8c99af3efa3278/hid.h#L176 within your Arduino code?

joseangeljimenez:
I finally could take some time to continue testing this. Here are my findings,

A) If I modify hidboot.h as you suggested, I get the desired results, perfectly receiving the fourth byte (wheel) in the HID report. Great!

B) However, if I call SetProtocol() after the initialization of the USB object, I don't receive the fourth byte. This is the code I am using,
```
static USB Usb;
static HIDBoot<HID_PROTOCOL_MOUSE> HidMouse(&Usb);
if (Usb.Init() == -1) {
[...]

PR #210, opened 2016-02-07
Added support to HIDBoot for easily enabling the Report Protocol

joseangeljimenez:
As per your suggestion in https://github.com/felis/USB_Host_Shield_2.0/issues/189, I have added support HIDBoot for easily enabling the Report Protocol.

Description:
"Added flag bRptProtoEnable to HIDBoot() constructor. The flag defaults to false in order to maintain current behavior (Boot Protocol). If flag is the to true, HIDBoot() will enable the Report Protocol. This modification is very useful for communication with HID devices with the full, richer, Report Protocol."

This is my first pull request to the library, your advice is welcome.

### DIFF
--- a/hidboot.h
+++ b/hidboot.h
@@ -210,6 +210,7 @@ class HIDBoot : public USBHID //public USBDeviceConfig, public UsbConfigXtracter
         uint32_t qNextPollTime; // next poll time
         bool bPollEnable; // poll enable flag
         uint8_t bInterval; // largest interval
+        bool bRptProtoEnable; // Report Protocol enable flag
 
         void Initialize();
 
@@ -218,7 +219,7 @@ class HIDBoot : public USBHID //public USBDeviceConfig, public UsbConfigXtracter
         };
 
 public:
-        HIDBoot(USB *p);
+        HIDBoot(USB *p, bool bRptProtoEnable=false);
 
         virtual bool SetReportParser(uint8_t id, HIDReportParser *prs) {
                 pRptParser[id] = prs;
@@ -252,10 +253,11 @@ public:
 };
 
 template <const uint8_t BOOT_PROTOCOL>
-HIDBoot<BOOT_PROTOCOL>::HIDBoot(USB *p) :
+HIDBoot<BOOT_PROTOCOL>::HIDBoot(USB *p, bool bRptProtoEnable) :
 USBHID(p),
 qNextPollTime(0),
-bPollEnable(false) {
+bPollEnable(false),
+bRptProtoEnable(bRptProtoEnable) {
         Initialize();
 
         for(int i = 0; i < epMUL(BOOT_PROTOCOL); i++) {
@@ -456,7 +458,7 @@ uint8_t HIDBoot<BOOT_PROTOCOL>::Init(uint8_t parent, uint8_t port, bool lowspeed
         // Yes, mouse wants SetProtocol and SetIdle too!
         for(uint8_t i = 0; i < epMUL(BOOT_PROTOCOL); i++) {
                 USBTRACE2("\r\nInterface:", i);
-                rcode = SetProtocol(i, USB_HID_BOOT_PROTOCOL);
+                rcode = SetProtocol(i, bRptProtoEnable ? HID_RPT_PROTOCOL : HID_BOOT_PROTOCOL);
                 if(rcode) goto FailSetProtocol;
                 USBTRACE2("PROTOCOL SET HID_BOOT rcode:", rcode);
                 rcode = SetIdle(i, 0, 0);
--- a/hidboot.h
+++ b/hidboot.h
@@ -219,7 +219,7 @@ class HIDBoot : public USBHID //public USBDeviceConfig, public UsbConfigXtracter
         };
 
 public:
-        HIDBoot(USB *p, bool bRptProtoEnable=false);
+        HIDBoot(USB *p, bool bRptProtoEnable = false);
 
         virtual bool SetReportParser(uint8_t id, HIDReportParser *prs) {
                 pRptParser[id] = prs;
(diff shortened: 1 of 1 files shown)


hidboot.h
```diff
@@ -456,7 +458,7 @@ uint8_t HIDBoot<BOOT_PROTOCOL>::Init(uint8_t parent, uint8_t port, bool lowspeed
         // Yes, mouse wants SetProtocol and SetIdle too!
         for(uint8_t i = 0; i < epMUL(BOOT_PROTOCOL); i++) {
                 USBTRACE2("\r\nInterface:", i);
-                rcode = SetProtocol(i, USB_HID_BOOT_PROTOCOL);
+                rcode = SetProtocol(i, bRptProtoEnable ? HID_RPT_PROTOCOL : HID_BOOT_PROTOCOL);
```
Lauszus: ```HID_BOOT_PROTOCOL``` should be ```USB_HID_BOOT_PROTOCOL```.

hidboot.h
```diff
@@ -252,10 +253,11 @@ class HIDBoot : public USBHID //public USBDeviceConfig, public UsbConfigXtracter
 };
 
 template <const uint8_t BOOT_PROTOCOL>
-HIDBoot<BOOT_PROTOCOL>::HIDBoot(USB *p) :
+HIDBoot<BOOT_PROTOCOL>::HIDBoot(USB *p, bool bRptProtoEnable) :
```
Lauszus: Please replace with:
```C++
HIDBoot<BOOT_PROTOCOL>::HIDBoot(USB *p, bool bRptProtoEnable/* = false*/) :
```

hidboot.h
```diff
@@ -218,7 +219,7 @@ class HIDBoot : public USBHID //public USBDeviceConfig, public UsbConfigXtracter
         };
 
 public:
-        HIDBoot(USB *p);
+        HIDBoot(USB *p, bool bRptProtoEnable=false);
```
Lauszus: Can you add a space before and after the equal? I.e:
```C++
HIDBoot(USB *p, bool bRptProtoEnable = false);
```

Lauszus:
@joseangeljimenez have you seen my inline comments?

joseangeljimenez:
Hello @Lauszus, sorry for the delay. I am right now making the changes...

joseangeljimenez:
This is the new commit,
https://github.com/felis/USB_Host_Shield_2.0/commit/a948b7afe5266331d937134d980d4aa91fc8cca0

Lauszus:
Thanks!

metaflow:
@joseangeljimenez Thank you for the fix! I have noticed that my mouse now reports `dx` and `dy` as int16_t instead of int8_t along with scroll (so full load is 7 bytes now:
```
buttons
dx low
dx high
dy low
dy high
wheel
unknown
```
Have you experienced that?

joseangeljimenez:
Yes. I am afraid that, when you activate the full HID report, the device is free to decide the format of the HID report. I tested a number of HID devices last year and found that there are different variations of the HID report.

You have two options here:

- If your device is "fixed" you can assume a specific full HID report. However, you will run into trouble if another HID device is used/plugged instead.
- You can write a small HID descriptor parser in order to read from the device the format of the full HID report. It is not that difficult, but may take you some days of coding and debugging.

xxxajk:
Parsing is actually quite easy to do, yes. 
Days of code? Not really.  Even if UHS 2 is pretty horrible in the parser department. The parser sort-of, but not quite follows the actual spec. Worse it does not return any of the information that you need for such things, only endpoint info. So yeah, you would have perhaps a day of work to get it going correctly.

I suggest that as soon as I work out the HID abstraction, that you move efforts to UHS 3.0.
Once you understand the simplicity of the API, making a new driver can happen in a matter of hours, not days. 

Here's the explanation of why you can manage to do it so quickly.

UHS 3 and beyond treats the entire USB descriptor as a file stream instead. 
This is way easier, since you just pick out the records you are interested in, and once you are done with the information in the record, you simply dispose of the allocated RAM, saving only the details that matter, which can usually end up a whole lot smaller than the descriptor format. Sometimes it can be reduced to a byte, which can be used as a flag that supports similar devices. 
Serial devices are a great example of this. There is only one interface for _ALL_ serial devices, and they emulate CDC-ACM, even if they are not CDC-ACM. 

One thing that I do want to stress here for UHS 3, which will make the experience wonderful for you, the end-user, is that no matter what kind of device you plug in, if there is a driver for it, the driver should convert everything and present it as a common abstraction, just like serial does. The idea here is that you should not have to write different code to support a different device in your sketch, which is ridiculous. When a new interface driver is implemented, the sketch does not need to be modified at all in order to use it. You just say in your code that you want to support it, and you are finished. It just works.

All UHS 3.x interface drivers also have the API designed that it is expandable, and modifiable in 2 key different ways. 

First,  if there is some feature to support that is useful, you can, and the code can fake it for devices that can't. Again, here old code which does not know about it, don't need modification, and once the new code becomes old, it will still work as expected.

Secondly, you can sub-class any interface driver in order to get it to do what you need. No more messing with the base source files. This can also be used in rapid similar interface driver development, and then applied to the base source once it is stabilized.

Lastly, and most important difference from previous implementations is that all UHS 3.x drivers are no longer "usb device drivers", instead they actually follow the spec and are proper interface drivers. This helps with devices that have multiple interfaces, as some could be already written, therefore will automatically support that part of the USB device. A perfect example here would be a USB device that has a serial interface, and also contains mass storage. If you were to design such a device, even if you emulated it with another board of your own design, it is supported without having to write an entire device driver like UHS1 and UHS2 requires. Another example is HID devices containing a hub. Adapters designed this way, such as some keyboard and mouse adapters, become supported out of the box automatically. Currently in UHS2 this is not the case, and never can be due to the design. It has to be told about the hub, and has to make special arrangements in order to use it. UHS3.x can enumerate such devices properly, and won't require any special parsers.

PR #214, opened 2016-02-25
Updated hub_demo to include string descriptors

stumpja:
The hub demo will now print usb string descriptors:

1] Manufacturer
2] Product Description
3] Serial Number

Also removed a new line return to make the output cleaner.

Please review.

### DIFF
--- a/examples/hub_demo/hub_demo.ino
+++ b/examples/hub_demo/hub_demo.ino
@@ -293,7 +293,7 @@ uint8_t getstrdescr( uint8_t addr, uint8_t idx )
     Serial.print("Error retrieving LangID table ");
     return ( rcode );
   }
-  langid = makeWord(buf[3], buf[2]);
+  langid = (buf[3] << 8) | buf[2];
   rcode = Usb.getStrDescr( addr, 0, 1, idx, langid, buf );
   if ( rcode ) {
     Serial.print("Error retrieving string length ");
@@ -418,10 +418,4 @@ void printProgStr(const char* str)
   if (!str) return;
   while ((c = pgm_read_byte(str++)))
     Serial.print(c);
-}
-
-/* Create a word from 2 bytes */
-uint16_t makeWord( uint8_t h, uint8_t l )
-{
-  return (h << 8) | l ;
 }
--- a/examples/hub_demo/hub_demo.ino
+++ b/examples/hub_demo/hub_demo.ino
@@ -93,7 +93,6 @@ void PrintAllDescriptors(UsbDevice *pdev)
   print_hex(pdev->address.devAddress, 8);
   Serial.println("\r\n--");
   getallstrdescr(pdev->address.devAddress);
-  //Serial.print("\r\n");
   PrintDescriptors( pdev->address.devAddress );
 }
 
--- a/examples/hub_demo/hub_demo.ino
+++ b/examples/hub_demo/hub_demo.ino
@@ -74,7 +74,7 @@ void PrintDescriptors(uint8_t addr)
   }
   Serial.print("\r\n");
 
-  for (int i = 0; i < num_conf; i++) {
+  for (uint8_t i = 0; i < num_conf; i++) {
     rcode = getconfdescr( addr, i );                 // get configuration descriptor
     if ( rcode ) {
       printProgStr(Gen_Error_str);
@@ -201,7 +201,7 @@ uint8_t getconfdescr( uint8_t addr, uint8_t conf )
   uint8_t rcode;
   uint8_t descr_length;
   uint8_t descr_type;
-  unsigned int total_length;
+  uint16_t total_length;
   rcode = Usb.getConfDescr( addr, 0, 4, conf, buf );  //get total length
   LOBYTE( total_length ) = buf[ 2 ];
   HIBYTE( total_length ) = buf[ 3 ];
@@ -275,13 +275,13 @@ uint8_t getallstrdescr(uint8_t addr)
 }
 
 //  function to get single string description
-unsigned int getstrdescr( unsigned int addr, uint8_t idx )
+uint8_t getstrdescr( uint8_t addr, uint8_t idx )
 {
   uint8_t buf[ 256 ];
-  unsigned int rcode;
+  uint8_t rcode;
   uint8_t length;
   uint8_t i;
-  unsigned short langid;
+  uint16_t langid;
   rcode = Usb.getStrDescr( addr, 0, 1, 0, 0, buf );  //get language table length
   if ( rcode ) {
     Serial.println("Error retrieving LangID table length");
@@ -293,7 +293,7 @@ unsigned int getstrdescr( unsigned int addr, uint8_t idx )
     Serial.print("Error retrieving LangID table ");
     return ( rcode );
   }
-  langid = word(buf[3], buf[2]);
+  langid = makeWord(buf[3], buf[2]);
   rcode = Usb.getStrDescr( addr, 0, 1, idx, langid, buf );
   if ( rcode ) {
     Serial.print("Error retrieving string length ");
@@ -411,7 +411,6 @@ void printunkdescr( uint8_t* descr_ptr )
   }
 }
 
-
 /* Print a string from Program Memory directly to save RAM */
 void printProgStr(const char* str)
 {
@@ -419,4 +418,10 @@ void printProgStr(const char* str)
   if (!str) return;
   while ((c = pgm_read_byte(str++)))
     Serial.print(c);
+}
+
+/* Create a word from 2 bytes */
+uint16_t makeWord( uint8_t h, uint8_t l )
+{
+  return (h << 8) | l ;
 }
--- a/examples/hub_demo/hub_demo.ino
+++ b/examples/hub_demo/hub_demo.ino
@@ -60,14 +60,14 @@ void setup()
   next_time = millis() + 10000;
 }
 
-byte getdevdescr( byte addr, byte &num_conf );
+uint8_t getdevdescr( uint8_t addr, uint8_t &num_conf );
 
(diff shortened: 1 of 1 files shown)


examples/hub_demo/hub_demo.ino
```diff
@@ -238,6 +239,83 @@ byte getconfdescr( byte addr, byte conf )
...
+ if( rcode ) {
+   Serial.print("Error retrieving string ");
+   return( rcode );
+ }
+ for( i = 2; i < length; i+=2 ) {
+   Serial.print((char) buf[i]);
+ }
+ return( rcode );
+}
+
```
Lauszus: Please fix the indentation in this functions. Btw the "Auto Format" tool in the Arduino IDE is actually quite good. Just run that and it should fix the issue :)

examples/hub_demo/hub_demo.ino
```diff
@@ -238,6 +239,83 @@ byte getconfdescr( byte addr, byte conf )
...
 }
+
+// function to get all string descriptors
+byte getallstrdescr(uint8_t addr)
+{ 
+  byte rcode;
+  Usb.Task();
+  if( Usb.getUsbTaskState() >= 0x80 ) {  // state configuring or higher
+    USB_DEVICE_DESCRIPTOR buf; 
+    rcode = Usb.getDevDescr( addr, 0, 0x12, ( uint8_t *)&buf );
```
Lauszus: Could you replace 0x12 with DEV_DESCR_LEN, so it will become clearer. There is one more here as well: https://github.com/stumpja/USB_Host_Shield_2.0/blob/master/examples/hub_demo/hub_demo.ino#L119. It would be nice if you could fix that as well while you at it.

examples/hub_demo/hub_demo.ino
```diff
@@ -238,6 +239,83 @@ byte getconfdescr( byte addr, byte conf )
   }//while( buf_ptr <=...
   return ( rcode );
 }
+
+// function to get all string descriptors
+byte getallstrdescr(uint8_t addr)
+{ 
+  byte rcode;
+  Usb.Task();
+  if( Usb.getUsbTaskState() >= 0x80 ) {  // state configuring or higher
```
Lauszus: 0x80 can also be replaced with ```USB_STATE_CONFIGURING```.

Lauszus:
Please see my inline comments. Also if you want it would be nice if all the ```byte``` types could be replaced with ```uint8_t```, as the ```byte``` type is deprecated.

examples/hub_demo/hub_demo.ino
```diff
@@ -238,6 +239,83 @@ byte getconfdescr( byte addr, byte conf )
...
+      }
+      Serial.print("\r\n");
+    }
+  }
+}
+
+//  function to get single string description
+unsigned int getstrdescr( unsigned int addr, byte idx )
+{
+ byte buf[ 66 ];
```
Lauszus: What is the reason for choosing this exact size? We might also want to add a check to make sure that the returned length does not exceed the buffer.

examples/hub_demo/hub_demo.ino
```diff
@@ -238,6 +239,83 @@ byte getconfdescr( byte addr, byte conf )
...
+   return( rcode );
+ }
+ length = buf[ 0 ];
+ rcode = Usb.getStrDescr( addr, 0, length, idx, langid, buf );
+ if( rcode ) {
+   Serial.print("Error retrieving string ");
+   return( rcode );
+ }
+ for( i = 2; i < length; i+=2 ) {
+   Serial.print((char) buf[i]);
```
Lauszus: Also a comment here that the string is unicode UTF-16 encoded might be useful, so people can figure out why we skip every second character.

stumpja:
@Lauszus I have made some minor changes, there are still unaddressed changes that need to be made. I will be looking into them later today.

Lauszus:
@stumpja thank you! Sorry for being so nitpicky, but I thought this would be a good time to fix some of the other issues in the example as well :)

stumpja:
@Lauszuz I never got around to cleaning up the remaining issues last Friday. However, I have addressed all your concerns mentioned above at this point. If I have not fixed one properly, or if you find something else, please let me know. I appreciate your "nitpicky-ness", honestly, and if we are going to update an example, we should do it right, so no worries there.

Your comment about the buffer size, this code was lifted from @felis  at the [circuits@home](https://www.circuitsathome.com/mcu/arduino-usb-host-part-3-descriptors) website. I never questioned it but after a little researching it seems that the maximum length of a string descriptor is 255 bytes since the first byte is reserved for the bLength variable. My points of reference can be found here:

[StackOverflow](http://stackoverflow.com/questions/7193645/how-long-is-the-string-of-manufacturer-of-a-usb-device)
[GitHub](https://github.com/nonolith/node-usb/issues/65)

They seem to agree so I updated the buffer length to 256 to accommodate the bLength variable and the UTF-16LE encoded string descriptor. Once we get this code where we want it, I will created another pull request updating the USB_desc example. Thanks for all your help with this!

examples/hub_demo/hub_demo.ino
```diff
@@ -238,6 +234,83 @@ byte getconfdescr( byte addr, byte conf )
...
+      Serial.print("\r\n");
+    }
+  }
+}
+
+//  function to get single string description
+unsigned int getstrdescr( unsigned int addr, uint8_t idx )
+{
+  uint8_t buf[ 256 ];
+  unsigned int rcode;
```
Lauszus: Please replace with ```uint8_t```

examples/hub_demo/hub_demo.ino
```diff
@@ -238,6 +234,83 @@ byte getconfdescr( byte addr, byte conf )
...
+  uint8_t length;
+  uint8_t i;
+  unsigned short langid;
```
Lauszus: Please replace with ```uint16_t```

examples/hub_demo/hub_demo.ino
```diff
@@ -238,6 +234,83 @@ byte getconfdescr( byte addr, byte conf )
...
+    Serial.println("Error retrieving LangID table length");
+    return ( rcode );
+  }
+  length = buf[ 0 ];      //length is the first byte
+  rcode = Usb.getStrDescr( addr, 0, length, 0, 0, buf );  //get language table
+  if ( rcode ) {
+    Serial.print("Error retrieving LangID table ");
+    return ( rcode );
+  }
+  langid = word(buf[3], buf[2]);
```
Lauszus: Please don't use the ```word``` function, as it is not standard and might cause troubles on non-Arduino boards.

examples/hub_demo/hub_demo.ino
```diff
@@ -198,13 +194,13 @@ void printhubdescr(uint8_t *descrptr, uint8_t addr)
...
 {
   uint8_t buf[ BUFSIZE ];
   uint8_t* buf_ptr = buf;
-  byte rcode;
-  byte descr_length;
-  byte descr_type;
+  uint8_t rcode;
+  uint8_t descr_length;
+  uint8_t descr_type;
   unsigned int total_length;
```
Lauszus: Can you replace this with ```uint16_t``` as well?

examples/hub_demo/hub_demo.ino
```diff
@@ -238,6 +234,83 @@ byte getconfdescr( byte addr, byte conf )
...
+      if ( rcode ) {
+        Serial.println( rcode, HEX );
+      }
+      Serial.print("\r\n");
+    }
+  }
+}
+
+//  function to get single string description
+unsigned int getstrdescr( unsigned int addr, uint8_t idx )
```
Lauszus: This should return an ```uint8_t``` and addr is just an ```uint8_t``` as well.

Lauszus:
@stumpja thanks for you hard work! I have added some other nitpicks, then it should be ready to merge :)

stumpja:
@Lauszus I've been preoccupied for a bit, but I have finally made the corrections mentioned above. Please look it over again.

Also, as far as I can tell the print_hex function is the only place where int's. Should we go ahead and swap these to uint8_t's as well?

I do apologize for this delay. Once we get this finalized, I will quickly update the USB_desc example.

Thanks!

examples/hub_demo/hub_demo.ino
```diff
@@ -293,7 +293,7 @@ unsigned int getstrdescr( unsigned int addr, uint8_t idx )
     Serial.print("Error retrieving LangID table ");
     return ( rcode );
   }
-  langid = word(buf[3], buf[2]);
+  langid = makeWord(buf[3], buf[2]);
```
Lauszus: Sorry, but don't use the makeWord function either, as this might not be available on non-Arduino boards. Simply just do it yourself:
```C++
(buf[3] << 8) | buf[2];
```

examples/hub_demo/hub_demo.ino
```diff
@@ -420,3 +419,9 @@ void printProgStr(const char* str)
   while ((c = pgm_read_byte(str++)))
     Serial.print(c);
 }
+
+/* Create a word from 2 bytes */
+uint16_t makeWord( uint8_t h, uint8_t l )
+{
+  return (h << 8) | l ;
+}
```
Lauszus: Please remove this, as this will cause conflicts when using Arduino: https://github.com/arduino/Arduino/search?utf8=%E2%9C%93&q=makeWord.

stumpja:
@Lauszus,

I'll eventually figure out what I am doing. Sorry. Update has been made.

Lauszus:
I just merged your PR manually: https://github.com/felis/USB_Host_Shield_2.0/commit/ef7d2e7bf5dda3051f68325e57653e307d3a90d6. Thank you :)

PR #220, opened 2016-03-21
Contribute MIDI host class driver

YuuichiAkagawa:
Added MIDI host class driver from https://github.com/YuuichiAkagawa/USBH_MIDI.
I contribute this library and join the UHS2.

### DIFF
--- a/examples/USBH_MIDI/USBH_MIDI_dump/USBH_MIDI_dump.ino
+++ b/examples/USBH_MIDI/USBH_MIDI_dump/USBH_MIDI_dump.ino
@@ -1,0 +1,98 @@
+/*
+ *******************************************************************************
+ * USB-MIDI dump utility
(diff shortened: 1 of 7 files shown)


Lauszus:
Thanks!

xxxajk:
How about a UHS 30 port??

YuuichiAkagawa:
@xxxajk Yes, I will.

xxxajk:
Awesomesauce!  Thank you! If you need any explanations to the new API ask
on the 3.0 issues list. Many things needed to be broken in order for things
to work better. I'm sure you will find the new API easier to work with, and
you should have actually less code to write. I recommend that if you are
going to do any enhancements to the driver or rename any of the methods in
the class that this is your grand opportunity to do so. UHS 3.0 is not
designed to be backwards compatible with 2.0 at all.

PR #227, opened 2016-04-16
Update avrpins.h

wilzzon17:
Good afternoon boy to develop a project with arduino mega adk , so I edit the library because it is necessary for my project.


Lauszus:
@wilzzon17 can you explain why this change would be needed? That lines makes sure that the Arduino ADK can be used directly with the library.

Lauszus:
I will close this PR, as this change really does not make any sense. Please reopen and explain in more detail if needed.

PR #237, opened 2016-05-01
Added missing function pointer to onInit on MultiWii example

cjsimon333:


Lauszus:
@cjsimon333 can you explain why this would be needed?

cjsimon333:
When compiling without the additional line of code, I get an error for the onInit argument not being defined. The Wii attachOnOnit method uses it as a function pointer, so I initialized it as a void pointer for the method. I assumed that others would also get this same error.

Lauszus:
Sorry but he onInit function is defined in the end of the example: https://github.com/felis/USB_Host_Shield_2.0/blob/master/examples/Bluetooth/WiiMulti/WiiMulti.ino#L125-L132. I just tried and the examples compiles fine here. What board are you using?

Lauszus:
@cjsimon333 you can also see from the Travis build that this change break the library: https://travis-ci.org/felis/USB_Host_Shield_2.0/builds/126961121.
I will close this PR now, as you commit does make much sense.

cjsimon333:
I'm using an Arduino Uno R3, using the Arduino IDE version 1.6.8. It gave me an error even with that onInit function definition on the bottom.

Lauszus:
Please post the error here.

cjsimon333:
```
Arduino: 1.6.8 (Windows 10), Board: "Arduino/Genuino Uno"

C:\Users\cjsimon333\Documents\Arduino\libraries\USB_Host_Shield_Library_2.0\examples\Bluetooth\WiiMulti\WiiMulti.ino: In function 'void setup()':

WiiMulti:29: error: 'onInit' was not declared in this scope

     Wii[i]->attachOnInit(onInit); // onInit() is called upon a new connection - you can call the function whatever you like
                          ^

exit status 1
'onInit' was not declared in this scope
```

YuuichiAkagawa:
Me too.
https://gist.github.com/YuuichiAkagawa/9f40cb23ee3ed50cf98d6905a53aaca1

Lauszus:
My guess would be that the function prototype is not being generated properly for some reason.
Please try to add:
```C++
void onInit();
```
At the top of the example and see if it compiles correctly.

Lauszus:
I can confirm that this is indeed an issue in Arduino 1.6.8 and have now been reported upstream: https://github.com/arduino/arduino-builder/issues/140. A workaround is to create the function prototype manually as per my previous comment.

PR #258, opened 2016-09-26
[RedBearLab] Fix type mismatch sizeof to uint8_t

sieren:
The RedbearLab compiler has a type mismatch for
STLs min() and sizeof. Thus we need to cast the
occurences of sizeof to uint8_t.

### DIFF
--- a/PS4Parser.cpp
+++ b/PS4Parser.cpp
@@ -84,7 +84,7 @@ void PS4Parser::Parse(uint8_t len, uint8_t *buf) {
 #endif
 
                 if (buf[0] == 0x01) // Check report ID
-                        memcpy(&ps4Data, buf + 1, min((uint8_t)(len - 1), sizeof(ps4Data)));
+                        memcpy(&ps4Data, buf + 1, min((uint8_t)(len - 1), MFK_CASTUINT8T sizeof(ps4Data)));
                 else if (buf[0] == 0x11) { // This report is send via Bluetooth, it has an offset of 2 compared to the USB data
                         if (len < 4) {
 #ifdef DEBUG_USB_HOST
@@ -93,7 +93,7 @@ void PS4Parser::Parse(uint8_t len, uint8_t *buf) {
 #endif
                                 return;
                         }
-                        memcpy(&ps4Data, buf + 3, min((uint8_t)(len - 3), sizeof(ps4Data)));
+                        memcpy(&ps4Data, buf + 3, min((uint8_t)(len - 3), MFK_CASTUINT8T sizeof(ps4Data)));
                 } else {
 #ifdef DEBUG_USB_HOST
                         Notify(PSTR("\r\nUnknown report id: "), 0x80);
--- a/PSBuzz.cpp
+++ b/PSBuzz.cpp
@@ -29,7 +29,7 @@ void PSBuzz::ParseHIDData(USBHID *hid, bool is_rpt_id, uint8_t len, uint8_t *buf
                         Notify(PSTR(" "), 0x80);
                 }
 #endif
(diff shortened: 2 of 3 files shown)


xxxajk:
Normally this should be size_t. 
Is this some kind of option passed to the compiler for using 8bit ints as a size optimization perhaps?
It also is doubtful that it is a compiler issue. It most likely is size optimizing library.
I recommend that you instead use a macro to work around the problem automatically.
Briefly:
1: Test if RedbearLab compiler, if not, set the macro to nothing
2: Test if it is this particular library and versions that will throw the error, if not set the macro to nothing
3: All tests passed, so set the macro to `(uint8_t)`

That would be a better/safer fix for everybody.

sieren:
How do I perform step 2? 
Testing for an error at compile-time then set the macro? Or do you expect me to go through all the library/compiler versions of the past months?

I'd assume users to use the most recent version, casting to `uint8_t` in case it's the RBL compiler should maintain backward compatibility anyway since we're using `uint8_t` on the 1st param as well, or am I wrong?

xxxajk:
If that is always the case with the particular library, then test if it is that specific lib. Normally there is a macro defined _someplace_.
For example I test for Newlib like so:
https://github.com/felis/UHS30/blob/master/libraries/UHS_host/mlock.c#L22
... I just don't care about the version, since it does not matter. If it doesn't matter for you, just do similar.

sieren:
I see, how do you reckon I should forego about the macro?
Simple redefining sizeof through a macro and adding the cast to it doesnt work (since there's places in the code where things are being added/subtracted and it becomes an int).
(Given by `work around automatically` you meant not touching any of the implementation code).
I could define some kind of casting macro and add it to the places as appropriate.

sieren:
```
.../USB_Host_Shield_2.0/PS4Parser.cpp:87:90: note:   deduced conflicting types for parameter 'const _Tp' ('unsigned char' and 'unsigned int')
                         memcpy(&ps4Data, buf + 1, min((uint8_t)(len - 1), sizeof(ps4Data)));
                                                                                          ^
../USB_Host_Shield_2.0/PS4Parser.cpp:96:90: error: no matching function for call to 'min(uint8_t, unsigned int)'
                         memcpy(&ps4Data, buf + 3, min((uint8_t)(len - 3), sizeof(ps4Data)));
```

xxxajk:
N00b QUESTION! :-)
You just roll your own.... 
```
#if defined(BLAH)
#define foo (uint8_t)
#else 
#define foo
#endif 

```
Just don't use 'foo', or you'll likely get shot ;-)

sieren:
Haha, yeah, well I know how to do THAT ;-)
But working on a foreign code-base I'm not sure what you folks use as best-practice here :)

xxxajk:
In UHS30 we prepend everything with UHS_.
As a rule that I use, when it is something like this in another project, prepend it with your initials and an underscore. For example I would do `AJK_FOO`...

xxxajk:
Oh yeah, and `STUPID_COMPILER` is already used IIRC.

sieren:
`STUPID_COMPILER` isn't very descriptive ;-)

Pushed a fix. I'd assume `settings.h` is the right place, since this is where I've done library specific changes the last time I had to fix something.

xxxajk:
Super, now only define it with the cast as that *IF* you are using your compiler. :-)
Then I think it will be dandy. :-)

xxxajk:
Ok, great. I'm hopeing that you are certain it won't affect someone using a different compiler and library on the same platform... we'll find out if anyone complains :-)

sieren:
That was a bit prematurely ;)
Haven't pushed the GCC Version check yet, though I guess there's only one compiler that's being used with RBL (at least thats what they distribute if you use their lib)

xxxajk:
That's fine, if we get a complaint, you can add that then.

PR #259, opened 2016-09-27
Add information about library to @PlatformIO Library Registry manifest file

Roman3349:
Add information about license and version to @PlatformIO Library Registry manifest file.

Signed-off-by: Roman3349 <ondracek.roman@centrum.cz>


Lauszus:
Thank you!

PR #266, opened 2016-11-02
Change INT pin for ESP8266 platform

huming2207:
Hi all,

Due to GPIO 6 to 11 are connected to internal SPI flash in the module and GPIO 16 cannot handle interrupts, we need to change the INT pin to another pin instead. In this occasion, I have changed it to GPIO 0. I haven't tested yet, but it should works.

See here for more details: #262 

Regards,
Jackson Hu.


Lauszus:
I will make this part of #263, so closing this for now.

PR #280, opened 2017-01-28
Address "dereferencing type-punned pointer" warnings.

arantius:
With compiler warnings set to "More", and compiling against this library, I see warnings of the form:

    hidboot.h:381:21: warning: dereferencing type-punned pointer will break strict-aliasing rules [-Wstrict-aliasing]
             num_of_conf = ((USB_DEVICE_DESCRIPTOR*)buf)->bNumConfigurations;
                         ^

This change addresses those warnings, making issues in my own code easier to notice.

### DIFF
--- a/hidboot.h
+++ b/hidboot.h
@@ -286,6 +286,7 @@ uint8_t HIDBoot<BOOT_PROTOCOL>::Init(uint8_t parent, uint8_t port, bool lowspeed
         const uint8_t constBufSize = sizeof (USB_DEVICE_DESCRIPTOR);
 
         uint8_t buf[constBufSize];
+        USB_DEVICE_DESCRIPTOR* device;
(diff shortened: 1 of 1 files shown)


Lauszus:
It would be better to fix this for the entire library. The following Travis build should show all the strict aliasing warnings: https://travis-ci.org/felis/USB_Host_Shield_2.0/builds/196773285.

Can you do a similar commit to this one: https://github.com/felis/USB_Host_Shield_2.0/commit/920f289a3f8288b28e11184cb0321b791e1fa5bc, so we can fix them all?

Lauszus:
It actually looks like it is only an issue with hidboot.h. I will merge your PR and then re-enable the warning for the Travis build.

Thanks!

Issue #273, opened 2016-12-19
PS4 slim controller does not connect via USB


kleindiek:
changed PID to 0x09cc (old PS4 controller has PID 0x05c4) in PS4USB.h

PS4 controller still does not connect.
PS4 controller works fine with BT.

#define ENABLE_UHS_DEBUGGING 1 in settings.h
Output:
PS4 USB Library StartedHU Init
Addr:01
NC:01

![img_4992](https://cloud.githubusercontent.com/assets/24652156/21313364/c9ae46ec-c5f1-11e6-8c72-bad24a5d2aa2.JPG)

Lauszus:
Could you run the following example: https://github.com/felis/USB_Host_Shield_2.0/blob/master/examples/USB_desc/USB_desc.ino?

kleindiek:
Here is the output:

Start


01
--

Device descriptor: 
Descriptor Length:	12
Descriptor type:	01
USB version:		0200
Device class:		00
Device Subclass:	00
Device Protocol:	00
Max.packet size:	40
Vendor  ID:		054C
Product ID:		09CC
Revision ID:		0100
Mfg.string index:	01
Prod.string index:	02
Serial number index:	00
Number of conf.:	01

Configuration descriptor:
Total length:		00E1
Num.intf:		04
Conf.value:		01
Conf.string:		00
Attr.:			C0
Max.pwr:		FA

Interface descriptor:
Intf.number:		00
Alt.:			00
Endpoints:		00
Intf. Class:		01
Intf. Subclass:		01
Intf. Protocol:		00
Intf.string:		00
Unknown descriptor:
Length:		0A
Type:		24
Contents:	01000147000201020C24
Unknown descriptor:
Length:		0C
Type:		24
Contents:	020101010602030000000A24
Unknown descriptor:
Length:		0A
Type:		24
Contents:	06020101030000000924
Unknown descriptor:
Length:		09
Type:		24
Contents:	030302040402000C24
Unknown descriptor:
Length:		0C
Type:		24
Contents:	020402040301000000000924
Unknown descriptor:
Length:		09
Type:		24
Contents:	060504010300000924
Unknown descriptor:
Length:		09
Type:		24
Contents:	030601010105000904

Interface descriptor:
Intf.number:		01
Alt.:			00
Endpoints:		00
Intf. Class:		01
Intf. Subclass:		02
Intf. Protocol:		00
Intf.string:		00

Interface descriptor:
Intf.number:		01
[...]

PR #281, opened 2017-01-30
Fixed bug for PS4 Slim controller via USB

Lauszus:
Use descriptor length in order to advance the buffer when parsing the descriptors.

This caused an issue with the PS4 Slim controller, as the endpoint descriptor was two bytes longer on the PS4 slim controller when reading the audio interface.

I suspect this is due to the fact that the audio interface (USB descriptor type 0x24) it is currently not supported by the library.

This is similar to what is done in the example as well: https://github.com/felis/USB_Host_Shield_2.0/blob/cbea36f76ffa0a0b1f74eb59e7ec845ccba380aa/examples/USB_desc/USB_desc.ino#L241.

Fixes issue #273.

### DIFF
--- a/PS4USB.h
+++ b/PS4USB.h
@@ -21,8 +21,9 @@
 #include "hiduniversal.h"
 #include "PS4Parser.h"
 
-#define PS4_VID 0x054C // Sony Corporation
-#define PS4_PID 0x05C4 // PS4 Controller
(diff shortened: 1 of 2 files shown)


Lauszus:
I also confirmed that the controller works via Bluetooth. If it is not working, then try another Bluetooth dongle.

PR #288, opened 2017-02-12
Fix missing unsigned comparison.

YuuichiAkagawa:
next_time value is ignored. This expression is always true.
```
hub_demo.ino:99:32: warning: comparison of unsigned expression >= 0 is always true [-Wtype-limits]

     if ((millis() - next_time) >= 0L) {

                                ^
```


Lauszus:
Thank you!

PR #298, opened 2017-06-15
Newer Xbox One controllers requires a longer initialisation command

### LINKED ISSUE #216: PS3USB and XBOXONE not working
IDE 1.6.7, UHS2 library 1.2.1

The discussion started on arduino.cc.

http://forum.arduino.cc/index.php?topic=382891.0

My PS3 controller connected via USB worked with an older version (1.0.0) of the library but does not work with 1.2.1. I found a workaround which involves commenting out 1 line in Usb.cpp. I am not sure this is the right way to fix the problem since this affects all USB devices. Debug output showed a stream of data toggle errors so I commented out the line that recovers from toggle errors. I can create a PR but this might break some other device.

```
Usb.cpp
@@ -234,7 +234,7 @@
                 rcode = dispatchPkt(tokIN, pep->epAddr, nak_limit); //IN packet to EP-'endpoint'. Function takes care of NAKS.
                 if(rcode == hrTOGERR) {
                         // yes, we flip it wrong here so that next time it is actually correct!
+                        // no, do not flip it because it generates more toggle errors
+                        //pep->bmRcvToggle = (regRd(rHRSL) & bmRCVTOGRD) ? 0 : 1;
-                        pep->bmRcvToggle = (regRd(rHRSL) & bmRCVTOGRD) ? 0 : 1;
                         regWr(rHCTL, (pep->bmRcvToggle) ? bmRCVTOG1 : bmRCVTOG0); //set toggle value
                         continue;
                 }

```

Another user has a similar problem with an XBOX ONE controller but the 1 line workaround does not help. I do not have an XBOX ONE controller so I can not do much more. I will post a link to this issue on the arduino.cc thread.

Lauszus:
Fixes #216

### DIFF
--- a/XBOXONE.cpp
+++ b/XBOXONE.cpp
@@ -179,7 +179,10 @@ uint8_t XBOXONE::Init(uint8_t parent, uint8_t port, bool lowspeed) {
         // initialize the controller for input
         writeBuf[0] = 0x05;
         writeBuf[1] = 0x20;
-        rcode = XboxCommand(writeBuf, 2);
+        writeBuf[2] = 0x00;
+        writeBuf[3] = 0x01;
+        writeBuf[4] = 0x00;
+        rcode = XboxCommand(writeBuf, 5);
         if (rcode)
                 goto Fail;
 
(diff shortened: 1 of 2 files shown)


Lauszus:
@guruthree could you test this on your old Xbox One controller?

PR #316, opened 2017-09-22
Make SPI port configurable for controllers with more than one

darkain:
Teensy micro controllers have more than one SPI port.

This simple change makes it easier to switch between SPI ports on these boards.

### DIFF
--- a/BTD.cpp
+++ b/BTD.cpp
@@ -418,7 +418,14 @@ void BTD::HCI_event_task() {
                                 break;
 
                         case EV_COMMAND_STATUS:
-                                if(hcibuf[2]) { // Show status on serial if not OK
+                                if(hcibuf[2] == 0x0B) {
+#ifdef DEBUG_USB_HOST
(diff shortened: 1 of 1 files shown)


settings.h
```diff
@@ -20,6 +20,12 @@ e-mail   :  support@circuitsathome.com
 #include "macros.h"
 
 ////////////////////////////////////////////////////////////////////////////////
+// SPI Configuration
+////////////////////////////////////////////////////////////////////////////////
+#define USB_SPI SPI
```
Lauszus: Please add ```ifndef``` around this. I.e:
```C++
#ifndef USB_SPI
#define USB_SPI SPI
#endif
```
This allows users to override it via a build flag.

Lauszus:
It seems like 6814e8afd336072a07b62ed85091074eddf4240f was pushed to this pull request as well by mistake.

Try to remove it, if you can't figure it out then I'll just do it when merging.

Lauszus:
Just merged the first commit manually in: #326. @darkain thanks for the contribution.

Issue #169, opened 2015-07-13
HCI Command Failed: 0B  - wiimote


moccault:
Hi,

I'm trying to connect my wiimote to arduino. I get this message:

![arduino wii](https://cloud.githubusercontent.com/assets/12648561/8655605/9624f548-2992-11e5-9c0a-37253ea8e58f.png)

Can't find the answer on the internet.

Thanks!

Michael

Lauszus:
Try to connect an external power supply, but to be honest it looks like a problem with your dongle.

moccault:
Hi,

Thanks for the fast reply!

I have 2 different bt dongles. They both work for the ps3 controller.
I powered my arduino with an external power supply.

Still I get the same message.

Regards

Michael

2015-07-13 23:11 GMT+02:00 Kristian Sloth Lauszus <notifications@github.com>
:

> Try to connect an external power supply, but to be honest it looks like a
> problem with your dongle.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/felis/USB_Host_Shield_2.0/issues/169#issuecomment-121060778>
> .
>



-- 
Met vriendelijke groeten

Michael Couckuyt

Lauszus:
Hmm it looks like the name is not read correctly. I believe https://github.com/felis/USB_Host_Shield_2.0/pull/168 fixes this, so can you download the newest version of the library and try again?

moccault:
Hi,

I only tested it with the latest version. To be sure I downloaded the
latest version again, still the same message.

Thanks,

Michael

2015-07-14 15:19 GMT+02:00 Kristian Sloth Lauszus <notifications@github.com>
:

> Hmm it looks like the name is not read correctly. I believe #168
> <https://github.com/felis/USB_Host_Shield_2.0/pull/168> fixes this, so
> can you download the newest version of the library and try again?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/felis/USB_Host_Shield_2.0/issues/169#issuecomment-121235584>
> .
>



-- 
Met vriendelijke groeten

Michael Couckuyt

Lauszus:
Okay. Can you try another dongle then?

HavenLau626:
Hi all,
I'm having the same issue as OP, 
I've tried powering the board with an external power supply and 3 different bluetooth dongles, I'm pretty sure the dongles work with my wiimote as I've had success with them pairing to my raspberry pi and reading values from it. 

[...]

PR #319, opened 2017-09-30
Fix for: HCI Command Failed: 0B

darkain:
This corrects the issue with the repeated 0x0B error messages when pairing a Wii Remote from #169

### DIFF
--- a/BTD.cpp
+++ b/BTD.cpp
@@ -418,7 +418,14 @@ void BTD::HCI_event_task() {
                                 break;
 
                         case EV_COMMAND_STATUS:
-                                if(hcibuf[2]) { // Show status on serial if not OK
+                                if(hcibuf[2] == 0x0B) {
+#ifdef DEBUG_USB_HOST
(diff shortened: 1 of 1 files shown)


Lauszus:
Thanks for the pull request. Could you implement this without using a delay? You can do this my extending the state machine.

The reason is that the library might be running on system where it is not allowed to block.

darkain:
The exiting stateCounter implementation is also broken though. It directly relies upon 1) the CPU clock speed, and 2) that the particular USB/BT/Wii Remote task is ran at a given interval. Time based code should be entirely re-written around something like elapsedMillis instead.

https://www.pjrc.com/teensy/td_timing_elaspedMillis.html
http://playground.arduino.cc/Code/ElapsedMillis
https://github.com/pfeerick/elapsedMillis/

Lauszus:
@darkain yes I agree. This is all taken care of in UHS30, as it is interrupt driven, so I think we should just leave the current state machine implementation as is.

Lauszus:
@darkain I have opened up a new PR. Please test #327 and report back if it works or not.

PR #343, opened 2018-01-18
Added boolean operator and more

neroroxxx:
Added a bool operator to check if a midi device is connected, added methods to get vid and pid to match those of Teensy's 3.6 built-in USB Host.

### DIFF
--- a/usbh_midi.h
+++ b/usbh_midi.h
@@ -66,6 +66,11 @@ protected:
 public:
         uint16_t pid, vid;
         USBH_MIDI(USB *p);
+        operator bool(){
+          return (pUsb->getUsbTaskState()==USB_STATE_RUNNING && isMidiFound);
+        }
+        uint16_t idVendor(){return vid;}
+        uint16_t idProduct(){return pid;}
         // Methods for recieving and sending data
         uint8_t RecvData(uint16_t *bytes_rcvd, uint8_t *dataptr);
         uint8_t RecvData(uint8_t *outBuf, bool isRaw=false);


Lauszus:
@YuuichiAkagawa can you take a look at this?

YuuichiAkagawa:
@Lauszus
Sure.

YuuichiAkagawa:
@neroroxxx 
Why do you want to use operator overloading?
Other drivers are not using it.

neroroxxx:
@YuuichiAkagawa it just seems like a clean way to use it, you can check if something is connected to the host AND if the device is a midi device in one statement.

Usb.Task();
if(Midi){
pollMIDI();
}

YuuichiAkagawa:
@neroroxxx But why only MIDI?

This code is pick from Teensy's library.
https://github.com/PaulStoffregen/USBHost_t36/blob/master/USBHost_t36.h#L388-L399

@Lauszus
Which do you think is better for applying to the whole UHS2 or only MIDI?

Lauszus:
@YuuichiAkagawa I don't see the benefit. I have only done it for the SPP class: https://github.com/felis/USB_Host_Shield_2.0/blob/master/SPP.h#L80-L82, as I wanted it to act much like the Serial class in Arduino.

usbh_midi.h
```diff
@@ -66,6 +66,11 @@ class USBH_MIDI : public USBDeviceConfig
 public:
         uint16_t pid, vid;
         USBH_MIDI(USB *p);
+        operator bool(){
+          return (pUsb->getUsbTaskState()==USB_STATE_RUNNING && isMidiFound);
+        }
+        uint16_t idVendor(){return vid;}
+        uint16_t idProduct(){return pid;}
```
Lauszus: @neroroxxx great if you could add some whitespace and use 8 spaces before the return statement, so it follows the coding style of the existing code.

YuuichiAkagawa:
@Lauszus OK. Please merge when the correction is completed.

Lauszus:
Merged manually in https://github.com/felis/USB_Host_Shield_2.0/commit/8ef3db254dcfa1fd27588c8401ac10cafe6c2bf4. Thanks for your contribution!

PR #365, opened 2018-03-15
Added support for rumble effect on Xbox One controllers

### LINKED ISSUE #363: Xbox One Rumble Support
Hi all,

I have been going through the library for the Xbox One controller and wanted to ask if code was already written to enable rumble. I see that there is code in the Xbox USB library and tried to replicate that in the Xbox One library to no avail. If there is support already available, how can I code it such that holding any of the trigger button will rumble the Xbox One controller? Any help would be great.

Thanks!

Lauszus:
Fixes #363


Lauszus:
@mojoe128 could you test this PR and report back if it works or not?

PR #409, opened 2018-09-19
Add SS, INT pins for Sanguino.

federicobriata:
Tested with USB Shield: DuinoFun UHS mini v2.0 (Aug. 16, 2014)


federicobriata:
Using Pin 10 and 9 it's a good choice for ATMega328, but on ATMega644p it's not the best option if you need to use UART0 and UART1.

Lauszus:
Okay makes sense :)

Issue #334, opened 2017-12-13
The HID device is not recognized through the USB hub.


XEROHN:
Hello!

I want to get the data of the keyboard and mouse through usb hub.
But I have a little problem.

I'm using the USBHIDBootKbdAndMouse.ino file.
`USBHub     Hub(&Usb);`
Of course I wrote the USBHub object.

The environment is the same as the picture.

![1](https://user-images.githubusercontent.com/31652986/33938366-9b1299d6-e04a-11e7-8f95-602a34e4c6c2.png)

connect a mouse or keyboard device (HID device) directly to the Shield USB port, data will be displayed.
But, if through the hub, only devices other than HID devices are recognized.

The serial output data is as follows.

```
Start
BM Init
Addr:0A
NC:01
HID_PROTOCOL_KEYBOARD AND MOUSE
bNumEP:01
D1
BM Init

getDevDescr 0D
BM Init

getDevDescr 0D
BM Init

getDevDescr 0D
BM Init

getDevDescr 0D

Lauszus:
@XEROHN from the picture it looks like you are using an unpowered USB hub. Can you try to connect an external power supply to the Arduino?

If that does not work, then I would recommend using a powered USB hub or connecting a 5V power supply directly to the Arduino, as all those USB device might draw a lot of current.

xxxajk:
Also, please note that this driver is for keyboard and mouse on a single device, such as a keyboard with built-in trackpad, or a usb to ps/2 convertor. 
For individual keyboard and mouse at the same time, you need to use the two separate drivers.

XEROHN:
@Lauszus 
I have tested with a total of six cases, Unfortunately, the output on the serial monitor did not change.

Jumperpad Soldered / Non-soldered
Powered Arduino / Usb hub / Arduino and Usb hub both

Behind the Usb hub shield is a jumper pad.
I've seen the circuit schematic and it seems to affect the power supply.
(On the schematic, the jumper pads(SJ1) are located near the Power pin socket and J2 pin socket.)
[https://www.arduino.cc/en/uploads/Main/USBHostShield-Rev2.pdf](url)

@xxxajk
As far as I know, the HID output data is parsing the data when the event occurs.
So, what do you think of running two Usb.Task in a threads?

Lauszus:
[...]

PR #414, opened 2018-10-11
fix some HUB related issues

ardi69:
1. fix Issue #334 in USB::AttemptConfig on failure the device is not correctly reseted because devConfig[parent] does not the right thing. Added function USB::ResetHubPort dues the job.
2. is a HUB is unplugged from a parent HUB the unplugged HUB is not released.
3. is a HUB released its childs are not released

### DIFF
--- a/usbhub.cpp
+++ b/usbhub.cpp
@@ -21,6 +21,7 @@ bool USBHub::bResetInitiated = false;
 USBHub::USBHub(USB *p) :
 pUsb(p),
 bAddress(0),
+bPortAddress(0),
 bNbrPorts(0),
 //bInitState(0),
 qNextPollTime(0),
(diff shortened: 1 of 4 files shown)


xxxajk:
Re #334 will test later today. Thank you!

ardi69:
i have forgotten: insert in usbhub.h
		`virtual uint8_t GetPortAddress() {
			return bPortAddress;
		};`

Lauszus:
@xxxajk did you ever test this?

Lauszus:
@xxxajk ping?

xxxajk:
Pong! will check today. Been overloaded with work...

gdsports:
PR #414 fixes problems I had with hub + USB mouse + USB keyboard using the hub_demo example. Looks fine.

The USBHIDBootKbdAndMouse example also works fine. Moving mouse and pressing keys results in serial console output as expected.

I will see if the same fixes apply to the SAMD port.

usbhub.cpp
```diff
@@ -104,6 +105,14 @@ uint8_t USBHub::Init(uint8_t parent, uint8_t port, bool lowspeed) {
         if(!bAddress)
                 return USB_ERROR_OUT_OF_ADDRESS_SPACE_IN_POOL;
 
+        {
+                UsbDeviceAddress a;
+                a.devAddress = bAddress;
+                a.bmHub = 0;
+                a.bmAddress = port ? : 1;
```
gdsports: Not sure what this does when port is non-zero. Is it the equivalent of
if (port) {
 a.bmAddress = ;
} else {
 a.bmAddress = 1;
}

xxxajk: Yes. Port numbers start at 1.
Right now commits are stopped, but eventually should be resumed.

PR #419, opened 2018-10-14
Fixes #417 - adds function gpioRdOutput to read output pins.

### LINKED ISSUE #417: Feature request: Ability to read the current GPIO outputs
The MAX3421E has the ability to read the current value of the GPIO output pins. It would be advantageous to expose this feature to through the library to allow an application to change the value of a single output without changing other outputs and removing the requirement to retain a record of the current output state.

From the datasheet:

> ...reading the state of a GPOUT7–GPOUT0 bit returns the state of the internal register bit, not the actual pin state. This is useful for doing read-modify-write operations to an output pin (such as blinking an LED), since the load on the output pin does not affect the register logic state.

riban-bw:


xxxajk:
Interesting idea. 
Does this actually report if an output pin is at the incorrect level, such as shorted high or low too, or just what was requested?

riban-bw:
This reports the state of the GPI outputs set by the MAX3421E. It does not read the actual level of the physical pin. So if you set the output high but external hardware drags it low, this function will still return a high value. From the datasheet:

> ..reading the state of a GPOUT7–GPOUT0 bit returns the state of the internal register bit, not the actual pin state.

xxxajk:
Just curious, but couldn't you hold a state variable for this?
I do understand it could be quite handy for verification and saving a byte of RAM on a small MCU.
I'm just wondering if the trade-off in speed and using code in flash is worth it.
I'm not knocking the patch, I'm just curious of your need for it.

riban-bw:
I considered holding the state in MCU memory to reduce the read-before-write time overhead but this may be corrupted if the MAX3421E is reset, i.e. the outputs may change value without updating the library's memory. The datasheet says that a chip reset does not change the GPI outputs but a power reset may occur for the MAX3421E whilst the MCU may remain powered in which case, assuming SPI comms continue, the two would be out of sync. If a user wants to avoid the read-before-write memory overhead then they could hold the value within user-space code.

xxxajk:
This gives me a nice idea... holding an 4-bit variable between reboots, where it doesn't have to be non-volatile. Just XOR the 4 higher bits to verify it is a real value... 
Could be handy for debug, even if I do have the AVR Dragon....
I like it!
I'll wait for @Lauszus to chime in with his thoughts, and if he does not respond within a few days with an objection, I'll accept the commit.

PR #421, opened 2018-10-14
Fixes #420 - Implements "Wiring" style simple access to GPIO interface

### LINKED ISSUE #420: Feature request: Implement "wiring" type GPIO access
It may be advantageous to implement simple digitalRead and digitalWrite functions to access individual GPIO pins. This would expose a familiar interface to user-space code and developers which may ease use of the GPIO interface.

I have a patch to implement this and will submit a pull request when it is tested.

riban-bw:


riban-bw:
Depends on PR #419

UHS2_gpio.cpp
```diff
@@ -0,0 +1,74 @@
...
+*/
+void UHS2_GPIO::digitalWrite(uint8_t pin, uint8_t val) {
+    if(pin > 7)
+        return;
+    uint8_t nValue = m_pUsb->gpioRdOutput();
+    uint8_t nMask = 1 << pin;
+    nValue &= (~nMask);
+    if(val)
+        nValue |= (nMask);
+    m_pUsb->gpioWr(nValue);
```
Lauszus: Please use an indentation of 8 spaces.

UHS2_gpio.h
```diff
@@ -0,0 +1,45 @@
...
+class UHS2_GPIO
+{
+    public:
+        UHS2_GPIO(USB *pUsb);
+
+        void digitalWrite(uint8_t pin, uint8_t val);
+        int digitalRead(uint8_t pin);
+        int digitalReadOutput(uint8_t pin);
+
+    private:
```
Lauszus: Please remove indentation in front of ```public``` and ```private```.

UHS2_gpio.h
```diff
@@ -0,0 +1,45 @@
...
+UHS2_GPIO implements "wiring" style GPIO access. Implemented by Brian Walton brian@riban.co.uk
+ */
+
+#if !defined(__USB2_GPIO_H__)
+#define __USB2_GPIO_H__
+
+#include "Usb.h"
+
+class UHS2_GPIO
+{
```
Lauszus: Please move ```{``` next to ```class UHS2_GPIO```.

examples/GPIO/Blink/Blink.ino
```diff
@@ -0,0 +1,40 @@
...
+
+// Satisfy the IDE, which needs to see the include statment in the ino too.
+#ifdef dobogusinclude
+#include <spi4teensy3.h>
+#endif
+
+#define OUTPUT_PIN 0
+
+USB Usb; //Create an UHS2 interface object
+UHS2_GPIO GPIO(&Usb); //Create a GPIO object
```
Lauszus: Please add space after ```//```.

examples/GPIO/Blink/Blink.ino
```diff
@@ -0,0 +1,40 @@
...
+  delay( 200 );
+}
+
+void loop()
+{
+  //Get the current output value, toggle then wait half a second
+  int nValue = GPIO.digitalReadOutput(OUTPUT_PIN);
+  nValue = (nValue?0:1);
+  GPIO.digitalWrite(OUTPUT_PIN, nValue);
+  Serial.print(nValue?"+":"."); //Debug to show what the output should be doing
```
Lauszus: Please replace this with:

```C++
Serial.write(nValue ? '+' : '.'); // Debug to show what the output should be doing
```

Same goes for the other example.

examples/GPIO/Blink/Blink.ino
```diff
@@ -0,0 +1,40 @@
...
+    Serial.println("OSC did not start.");
+
+  delay( 200 );
+}
+
+void loop()
+{
+  //Get the current output value, toggle then wait half a second
+  int nValue = GPIO.digitalReadOutput(OUTPUT_PIN);
+  nValue = (nValue?0:1);
```
Lauszus: Maybe we should replace this with:

```C++
nValue = nValue ? LOW : HIGH;
```

So it is a little easier for beginners.

examples/GPIO/Blink/Blink.ino
```diff
@@ -0,0 +1,40 @@
...
+  if (Usb.Init() == -1)
+    Serial.println("OSC did not start.");
+
+  delay( 200 );
+}
+
+void loop()
+{
+  //Get the current output value, toggle then wait half a second
+  int nValue = GPIO.digitalReadOutput(OUTPUT_PIN);
```
Lauszus: Please use ```uint8_t``` instead of ```int```.

examples/GPIO/Blink_LowLevel/Blink_LowLevel.ino
```diff
@@ -0,0 +1,40 @@
...
+    Serial.println("OSC did not start.");
+
+  delay( 200 );
+}
+
+void loop()
+{
+  //Get the current output value, toggle then wait half a second
+  uint8_t nGPO = Usb.gpioRdOutput();
+  uint8_t nValue = ((nGPO & 0x01) == 0x01)?0:1;
```
Lauszus: Please add some spaces here.

examples/GPIO/Blink_LowLevel/Blink_LowLevel.ino
```diff
@@ -0,0 +1,40 @@
...
+  nGPO &= 0xFE; //clear bit 0
+  nGPO |= nValue;
+  Usb.gpioWr(nGPO);
+  Serial.print(nValue?"+":"."); //Debug to show what the output should be doing
```
Lauszus: See previous comment.

Lauszus:
Also could you please merge in master, as I have now merged #419.

Lauszus:
Could you please also add the examples to Travis: https://github.com/felis/USB_Host_Shield_2.0/blob/master/.travis.yml?

UHS2_gpio.cpp
riban-bw: I don't get this! Do you want the whole function indented by 8 spaces (which seems excessive indentation) or just these four lines (which doesn't make sense to me as these are in same block as rest of function).

examples/GPIO/Blink/Blink.ino
riban-bw: The function returns int (as per the "Wiring" digitalRead). I agree this is wrong and would personally have implemented this as returning bool but I tried to be consistent with the Arduino way, which is what this PR is about. The example uses the same data type as the digitalReadOutput method which is int. The example should be consistent with the library.

xxxajk: I agree, this is OK.

UHS2_gpio.cpp
```diff
@@ -0,0 +1,74 @@
...
+UHS2_GPIO implements "wiring" style GPIO access. Implemented by Brian Walton brian@riban.co.uk
+ */
+
+#include "UHS2_gpio.h"
+
+/** @brief  Implement an instance of a UHS2_GPIO object
+*   @param  pUSB Pointer to a UHS2 USB object
+*/
+UHS2_GPIO::UHS2_GPIO(USB *pUsb) : m_pUsb(pUsb)
+{
```
xxxajk: Please cuddle the brace.

examples/GPIO/Blink/Blink.ino
```diff
@@ -0,0 +1,40 @@
...
+#include <spi4teensy3.h>
+#endif
+
+#define OUTPUT_PIN 0
+
+USB Usb; // Create an UHS2 interface object
+UHS2_GPIO GPIO(&Usb); // Create a GPIO object
+
+void setup()
+{
```
xxxajk: Same here, cuddle all braces. This isn't 1979 C.

examples/GPIO/Blink/Blink.ino
xxxajk: One thing, indents are 8 spaces for a very specific set of reasons.
First, simply makes code easier for a human to read, which is seriously important.
Second, looks non-cluttered, and consistent across the entire library.
Finally, 8 space tabs have been a real actual standard since forever. 

Just because a few people think different, does not make it correct, especially for the above reasons.
If your text editor can accept better indents, and perform code formatting, please use those features.
Also, try to ensure that you don't have any trailing whitespace, for all the obvious reasons.

UHS2_gpio.cpp
riban-bw: Fixed in commit 5046bc8.

examples/GPIO/Blink/Blink.ino
riban-bw: Fixed in commit 5046bc8. To cuddle or not to cuddle isn't really a matter age but a coding style decision. Many projects I work on use uncuddled braces (and is my preference) but I did try to comply with the coding style I inferred from other software modules in this project. The example was based on the MIDI examples which do not cuddle. I did try! Anyway - sorted now and I will try to do better in the future :-).

xxxajk: Not a problem :-)
Tab spacing has always been an argument, which I won't hash out here, but K&R were right on the 8 spaces, and besides, that's the default on any computer's text output and the old typewriters and printers, etc. That plus years of coding Linux Kernel have paid off stylistically.
You get a cookie for at least trying, and a beer for the patch as a whole.

examples/GPIO/Blink/Blink.ino
riban-bw: The great thing about standards is that there are so many to choose from. Coding styles are generally chosen by a project (or, so often not chosen which is a pain) based on the lead's personal preference. I tend to use Allman (ANSI) coding style for C / C++ projects and use the Code::Blocks IDE source formatter plugin to support this style. So I generally don't cuddle braces and use 4 spaces indents which I find sufficiently clear without pushing long code lines off the screen / onto next line. When writing Java I will use the Java style and when contributing to other projects will use whatever style is either mandated / suggested or inferred from existing code (as I did for this PR). The examples in this project are a little inconsistent so may benefit from rationalising to meet your coding style. I am not aware of 8 spaces being a "real actual standard" and believe that coding style is implemented by guidelines. Please point me at any actual published standards as I have been involved in coding style guidelines for many years and would be interested to see where standards have been set. (It is many years (but not 1970's) since I learned to code and it is easy for an old man to miss the new big thing!) Cheers!

Lauszus: I see your point with int, so let's stick with that.

I personally are not a big fan of 8 spaces, but that was decided a long time ago, so now we have to stick with it. And no there is no such thing as a coding style standard for C/C++ like they have in Python.

And yes some of the old examples are a little inconsistent.

PR #430, opened 2018-11-21
Add PID for Xbox Adaptive Controller

gdsports:
[XAC](https://www.xbox.com/en-US/xbox-one/accessories/controllers/xbox-adaptive-controller) works just like an Xbox One but the PID is new.


Lauszus:
Thanks!

PR #431, opened 2018-11-24
PS4 PAIRING TROUBLE

chBSS:
Dear All
I try to pair my PS4 dualshock with an Arduino DUE. I Used  PS4BT.ino including in the "USB_Host_Shield_2.0-master"
I use a dongle TRENDnet TBW 106UB as indicated in the dongle compatible list. I addition, I tested several other dongle, (in case of), Always the same result.

When PS4BT.ino lauched.
I pressed hold PS and Share button at the same time, the white led blink 2 times, Off and 2 times...etc

In the serial monitor I can read the folliwing lines.  … all seems good… but dualshock does not work. 

I need help to resolve / understand the problem .. by advance many thanks

christopher

**SERIAL MONITOR MESSAGE**

PS4 Bluetooth Library Started
Bluetooth Dongle Initialized
No response to HCI Reset
HCI Reset complete
Write class of device
Local Bluetooth Address: 00:15:83:F3:56:B0
Please enable discovery of your device

Gamepad found
HID device found
Now just create the instance like so:
BTHID bthid(&Btd);
And then press any button on the device
Connecting to HID device
Connected to HID device
Received Key Request
Bluetooth pin is set too: 0000
Pairing successful with HID device
Send HID Control Connection Request


Lauszus:
Seems like this is really an issue and not PR. Please reopen a new issue if needed.

PR #438, opened 2019-01-08
Fix some bugs

gyojir:
1. Do not reset Global Items rptSize and rptCount on Main Items
In Report Descriptor parsing, Global Item will remain after the Main Item.

2. Do not flip toggle value on Toggle Error
In [MAX3421E Programming Guide ](https://pdfserv.maximintegrated.com/en/an/AN3785.pdf) p13 

> By ACK-ing the transfer and not updating its own toggle bit, the SIE causes the peripheral to complement its toggle bit, thus forcing the data toggle mechanism back into sync. 

In short, there is no need to flip on a toggle error.
With this fix my Logitech Gamepad F310 is now working.

3. Fix reading of data with offset
Reading of multiple bytes of data probably got bugged, so I fixed it.

### DIFF
--- a/Usb.cpp
+++ b/Usb.cpp
@@ -172,8 +172,6 @@ uint8_t USB::ctrlReq(uint8_t addr, uint8_t ep, uint8_t bmReqType, uint8_t bReque
 
                                 rcode = InTransfer(pep, nak_limit, &read, dataptr);
                                 if(rcode == hrTOGERR) {
-                                        // yes, we flip it wrong here so that next time it is actually correct!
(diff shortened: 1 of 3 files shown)


Lauszus:
@felis since this is part of the code it would be good if you could verify this?

doctea:
This seems to solve the problem I mention in https://github.com/YuuichiAkagawa/USBH_MIDI/issues/73# that prevented MIDI data and MIDI clock from working on the Arturia Beatstep!

PR #453, opened 2019-03-10
Add support for Thrustmaster T.16000M FCS joystick

gdsports:


Lauszus:
Thank you for the PR :)

BlueGene00:
Hi,
I have tested this example with the T.16000M and it works.

Now I want the process the values, but I am not quite sure how to this.

The example has:
	```
Serial.print("X: ");
	PrintHex<uint16_t>(evt->x, 0x80);
	Serial.print(" Y: ");
	PrintHex<uint16_t>(evt->y, 0x80);
```

How can I convert those HEX numbers and use them with an integer variable?

**Comment by gdsports** [comment]
evt->x is an unsigned 16 bit integer. You can replace all of the above with the following.

```
Serial.println(evt->x);
Serial.println(evt->y);
```

BlueGene00:
Thanks, it worked.

The next step it to read the buttons. One button at a time is no problem. But what about multiply buttons at the same time? 

I know it is not quite common, but would make the implementation complete. As far as I saw, the values are added together. Are to any functions to differentiate between multiply buttons at the same time?

PR #467, opened 2019-06-19
Typo

njwhite:


Lauszus:
@njwhite thanks. I found the same typo in several of the examples, but should be fixed in: https://github.com/felis/USB_Host_Shield_2.0/commit/49421b6ce40442d5afe3097ab0f3e65b8fc69aa5.

PR #470, opened 2019-06-28
Fix NAK data corruption

gdsports:
See https://github.com/felis/USB_Host_Shield_2.0/issues/403

PR #473, opened 2019-07-02
Fix corruption. NEVER use SPI method transfer on a pointer.

xxxajk:
Should fix the problems.
Also removes duplicate code.

### DIFF
--- a/usbhost.h
+++ b/usbhost.h
@@ -228,21 +228,15 @@ uint8_t* MAX3421e< SPI_SS, INTR >::bytesWr(uint8_t reg, uint8_t nbytes, uint8_t*
         spi4teensy3::send(reg | 0x02);
         spi4teensy3::send(data_p, nbytes);
         data_p += nbytes;
-#elif defined(SPI_HAS_TRANSACTION) && !defined(ESP8266) && !defined(ESP32)
-        USB_SPI.transfer(reg | 0x02);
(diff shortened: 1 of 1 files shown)


gdsports:
The USBPrinter_uhs2 driver works with this patch using Uno and Leonardo. I do not have a Mega 2560 but the patch should work. Also the printer is confirmed to work on ESP8266. I will withdraw PR #470 since it is redundant.

BTW, M5Stack released a USB host MAX3421E module so I will be testing this out when I receive one. M5Stack is an ESP32 based system of stackable modules. The modules are like shields but include plastic casing so the module stack is professional looking.

https://m5stack.com/collections/m5-module/products/usb-module

Lauszus:
Is this ready to merge?

gdsports:
It has been working fine for me.

Lauszus:
Thanks for the report @gdsports. I'll make a new release as well, so all the people that are using the Arduino library manager will get the update as well.

Bacto:
Hi everyone ant thanks for this library :)
Do you have any news about the next release?
The last one is from 2018/15/03 and does not include this fix unfortunately :-/

Vamshi253:
Hi @gdsports,

The USBPrinter_uhs2 driver is not working with this patch using Arduino Mega ADK. Could you please suggest any changes that I need to make in order to work out with Arduino Mega ADK.

Thank you

PR #517, opened 2020-02-28
Add a ParseBTHIDControlData virtual method similar to ParseBTHIDData,

fraca7:
for reports that are sent through the control pipe. This happens for
instance in response to a GET_REPORT.


Lauszus:
Thanks for the PR.

PR #522, opened 2020-03-26
hiduniversal don't ignore same hiddata

liumazi:
Only the first mouse wheel event can be received, if discard the same hiddata.
BTW: A Curious Thing, if annotated “ZeroMemory(constBuffLen, prevBuf);”, it can't poll any data ?


quex46:
Duplicate of #510 

I've just hit same issue with mouse scrolls and smooth mouse pointer movements (most of them are skipped).

Lauszus:
Fixed in https://github.com/felis/USB_Host_Shield_2.0/pull/584.

PR #528, opened 2020-04-25
Added new BulkOnly public functions: SCSITransaction12() and SCSITransaction16()

adamm:
This will allow for compatibility with devices requiring 12-byte and 16-byte CDB transactions, and for support for 32-bit and 64-bit LBAs.


Lauszus:
Sorry for not getting back to you. I can really merge it as is because you changed the INT pin.

adamm:
Hi,

I cannot remember why I did this, but there was probably a good reason :)
Let me look into it and resubmit.
-- 
Adam McDaniel

PR #531, opened 2020-05-03
Flush the serial buffer

bartstasik:
With enough hammering, the buffer gets filled with more data than it can
handle and causes problems with the MIDI output. Ideally, there 
shouldn't be a buffer here in the first place, because we're dealing 
with constant time data anyway.

Flush the buffer after sending the USB host data to the serial port.

### DIFF
--- a/examples/USBH_MIDI/USB_MIDI_converter/USB_MIDI_converter.ino
+++ b/examples/USBH_MIDI/USB_MIDI_converter/USB_MIDI_converter.ino
@@ -68,6 +68,7 @@ void MIDI_poll()
     if ( (size = Midi.RecvData(outBuf)) > 0 ) {
       //MIDI Output
       _MIDI_SERIAL_PORT.write(outBuf, size);
+      _MIDI_SERIAL_PORT.flush()
     }
   } while (size > 0);
 }


Lauszus:
@YuuichiAkagawa please check this.

YuuichiAkagawa:
Umm, this is just example.
It is necessary or not is case by case.
If you need it, add it to your own sketch.

bartstasik:
> Umm, this is just example.
> It is necessary or not is case by case.
> If you need it, add it to your own sketch.

Sure, it's just an example, but there's no need to keep the data in the buffer once it's been sent off. The way I understand an example is that it is as best of a representation of something as we can get. Someone can just take things away if they don't need it :)

YuuichiAkagawa:
> Someone can just take things away if they don't need it :)

and vice versa.
I disagree with this change.

Lauszus:
Maybe we could keep the code, but have it commented out and then have a comment above explaining why people might want to use it?

YuuichiAkagawa:
Like this?
```
      //If you want send data immediately, uncommenting the following line.
      //_MIDI_SERIAL_PORT.flush();
```

Don't forget the semicolon.

Lauszus:
@YuuichiAkagawa yes exactly! @bartstasik is that okay with you?

bartstasik:
@Lauszus Up to you! :) I suppose it's important to show that this exists even if the line isn't actually being used, but at that point, I would perhaps suggest documenting this properly instead of adding clutter into the code.

Correct me if I'm wrong though: flushing the buffer after writing to the serial port gets rid of the data that has already been sent. Without flushing, the data is still held in the buffer and so it would eventually become full, thus (eventually) resulting in delayed transmission. With MIDI, unexpected (delayed) signals sound to me like the last thing someone would want, because it could ruin the performance.

 I'm more than happy to be corrected and I welcome learning whenever possible :)

Lauszus:
@bartstasik I would say the best way to document is to put it in the example code. You can also add it to the following section in the README: https://github.com/felis/USB_Host_Shield_2.0#midi-library, but in my experience most people dot not read the README (just check all the issues).

> Correct me if I'm wrong though: flushing the buffer after writing to the serial port gets rid of the data that has already been sent. Without flushing, the data is still held in the buffer and so it would eventually become full, thus (eventually) resulting in delayed transmission. With MIDI, unexpected (delayed) signals sound to me like the last thing someone would want, because it could ruin the performance.

Since @YuuichiAkagawa wrote the code it's his decision. I think documenting it in the code and README is a good compromise :)

YuuichiAkagawa:
@Lauszus Shall we create new issue and pull request? issue is this discussion. pull request is commented code with link for issue. 
```
      //https://github.com/felis/USB_Host_Shield_2.0/issues/nnn
      //If you want to send data immediately, uncommenting the following line.
      //_MIDI_SERIAL_PORT.flush();
```
@bartstasik What Serial.flush() provides is to wait for the transmission to complete.
If you write data without flashing when there is a large amount of data from USB, the overflowed data may be lost. It's not a delay. 
As a performance issue, waiting for the flush to complete delays reading from USB.

Lauszus:
@YuuichiAkagawa yes that would be great!

Lauszus:
@YuuichiAkagawa any change you can open a new PR with your suggested change:

> ```
>       //https://github.com/felis/USB_Host_Shield_2.0/issues/nnn
>       //If you want to send data immediately, uncommenting the following line.
>       //_MIDI_SERIAL_PORT.flush();
> ```

YuuichiAkagawa:
Oops. I completely forgot about it. I'll do it later.

Lauszus:
@YuuichiAkagawa thanks!

Lauszus:
Fixed in https://github.com/felis/USB_Host_Shield_2.0/pull/571.

PR #532, opened 2020-05-15
fix gcc warnings

doj:
these changes fix warnings with the latest Arduino IDE version.
- ignore warnings about case fall through.
- fix warning about unused parameter. I've opted for the "(void)var;" way, as I find it better than cluttering the parameter definition with the attribute unused.


Lauszus:
@doj thanks for the PR! I fixed this and some other stuff needed for the Travis build to succeeded in this PR: https://github.com/felis/USB_Host_Shield_2.0/pull/471.

Issue #570, opened 2020-12-01
Flush the serial buffer in MIDI


YuuichiAkagawa:
This discussion has been move from #531.

Serial.write() stores data in a buffer and sends it asynchronously.

You can wait for the transmission to complete by calling Serial.flush().
It does not send the data in the buffer immediately. Just wait.
The wait time is about 1 ms per MIDI message.

You don't necessarily have to call Serial.flush (), but when processing large amounts of data such as Sysex, the send buffer can overflow. In this case it may be better to wait for the transmission to complete.

We plan to make the following modifications to the MIDI examples so that you can choose whether or not to use the flash. 

```
// Set to 1 if you want to wait for the Serial MIDI transmission to complete.
#define ENABLE_MIDI_SERIAL_FLUSH 0
.
.
.
#if ENABLE_MIDI_SERIAL_FLUSH
      _MIDI_SERIAL_PORT.flush();
#endif
```

Lauszus:
Fixed in https://github.com/felis/USB_Host_Shield_2.0/pull/571.

PR #571, opened 2020-12-01
Add Serial.flush() option for MIDI examples

YuuichiAkagawa:
PR for #570 
Add Serial.flush() option for MIDI examples.


Lauszus:
Thanks!

PR #574, opened 2020-12-08
Add Xbox Core Controller PID

dmadison:
Adds the PID for the new Xbox Series S / Series X "core" controllers to the `XBOXONE` wired USB class. Tested and working.

These controllers also include a new 'share' button, which is bit 0x01 of byte 22 in the report. Although since the `uint16_t` button variable is already full I'm not sure how best to add it to the class.


Lauszus:
Thank you for the PR! Feel free to send another one adding support for the share button.

dmadison:
Do you have any advice on what to name the enum entry? It looks as though there's a conflict, as `SHARE` is already used by the PS4 definitions and has the same value as `BACK` / `VIEW`.

PR #583, opened 2021-01-12
HIDUniversal, HIDComposite: Don't overflow hidInterfaces[] or epInfo[]

DanielGibson:
If a connected device has more than 3 (maxHidInterfaces) HID interfaces,
which is not unusual with modern keyboards, EndpointXtract() wrote
beyond the hidInterfaces[] array and corrupted bHasReportId, PID + VID.

The same could happen with the epInfo[] array.
Now this is fixed by checking bNumIface/bNumEP before adding new
elements to those arrays.

### DIFF
--- a/hidcomposite.cpp
+++ b/hidcomposite.cpp
@@ -306,6 +306,8 @@ void HIDComposite::EndpointXtract(uint8_t conf, uint8_t iface, uint8_t alt, uint
 
         // Fill in interface structure in case of new interface
         if(!piface) {
+                if(bNumIface >= maxHidInterfaces)
+                        return; // don't overflow hidInterfaces[]
                 piface = hidInterfaces + bNumIface;
                 piface->bmInterface = iface;
                 piface->bmAltSet = alt;
@@ -319,7 +321,7 @@ void HIDComposite::EndpointXtract(uint8_t conf, uint8_t iface, uint8_t alt, uint
         if(!SelectInterface(iface, proto))
                 index = 0;
 
-        if(index) {
+        if(index && bNumEP < totalEndpoints) {
                 // Fill in the endpoint info structure
                 epInfo[bNumEP].epAddr = (pep->bEndpointAddress & 0x0F);
                 epInfo[bNumEP].maxPktSize = (uint8_t)pep->wMaxPacketSize;
--- a/hidcomposite.cpp
+++ b/hidcomposite.cpp
@@ -306,8 +306,13 @@ void HIDComposite::EndpointXtract(uint8_t conf, uint8_t iface, uint8_t alt, uint
 
         // Fill in interface structure in case of new interface
         if(!piface) {
-                if(bNumIface >= maxHidInterfaces)
-                        return; // don't overflow hidInterfaces[]
+                if(bNumIface >= maxHidInterfaces) {
+                        // don't overflow hidInterfaces[]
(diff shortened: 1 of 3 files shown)


DanielGibson:
I wish I had found this before the release a few days ago..

As a side-note, I wonder why `HIDUniversal` and `HIDComposite` are separate classes, they're almost identical?
Furthermore, `bHasReportId` is more or less unused: It's only set once, to `false`, in the constructors. And it doesn't make much sense at class level anyway, as it can be different per endpoint - see for example multimedia keyboards, which usually have one interface with one endpoint for the USB boot-protocol keyboard part (which does *not* use Report IDs) and one interface with one endpoint for the consumer control and system control ("multimedia keys") part (and possibly additional things like extra NKRO reports) that *does* use Report IDs.

Lauszus:
Good find! 

Could you add some serial debug code, so the user gets notified if the number of interfaces or endpoints are exceeded? (See for example these lines on how to add debug output to the library: https://github.com/felis/USB_Host_Shield_2.0/blob/master/BTHID.cpp#L89-L102).

> I wish I had found this before the release a few days ago..

No problem. I'll just make a new one :)

> As a side-note, I wonder why HIDUniversal and HIDComposite are separate classes, they're almost identical?

Maybe create a base class and then have the `HIDUniversal` and `HIDComposite` inherit it? PR*s are more than welcome :)

I guess `bHasReportId` was included by felis originally in case the code got extended. Feel free to remove it.

DanielGibson:
Ok, I'll add the debug output and I'll try to unify the classes with inheritance.

Do you think it'd cause problems if I increased `maxHidInterfaces` a bit, maybe to 4 or 5?

Lauszus:
> Ok, I'll add the debug output and I'll try to unify the classes with inheritance

Thanks! Could you implement this PR into it as well: https://github.com/felis/USB_Host_Shield_2.0/pull/510, as I never got around to merging it...

> Do you think it'd cause problems if I increased maxHidInterfaces a bit, maybe to 4 or 5?

No that is alright. We can use the CI build to check the RAM and Flash usage of an Arduino Uno to make sure the examples do not get too close to the limit.

DanielGibson:
I increased the supported number of interfaces and added log messages to the overflow checks.

I think I'll make a separate PR for unifying HIDUniversal and HIDComposite (and possibly removing `bHasReportId`) that will also include the changes of #510 (if I get it all to work).

Lauszus:
Thanks. I checked the RAM and Flash usage for the `USBHIDMultimediaKbd.ino` example on an Arduino Uno.

Before:

```
RAM:   [==        ]  25.0% (used 512 bytes from 2048 bytes)
Flash: [===       ]  33.7% (used 10870 bytes from 32256 bytes)
```

After:

```
RAM:   [===       ]  25.9% (used 530 bytes from 2048 bytes)
Flash: [====      ]  35.4% (used 11418 bytes from 32256 bytes)
```

As you can see it's a very minimal increase, so I do not think increasing `maxHidInterfaces` to 5 will cause any issues.

Thus I will merge this PR.

Lauszus:
> I think I'll make a separate PR for unifying HIDUniversal and HIDComposite (and possibly removing bHasReportId) that will also include the changes of #510 (if I get it all to work).

Great. Looking forward to the next PR :)

DanielGibson:
Thanks!

BTW, on an unrelated note, http://shop.tkjelectronics.dk/ is down (I assume you know whom to ping about that?)

Lauszus:
Thanks. It's up and running again now :)

PR #584, opened 2021-01-13
Unify HIDComposite and HIDUniversal

DanielGibson:
* Integrated the change from #510 and cleaned it up (removed things not needed anymore due to that change)
* HIDUniversal is now derived from HIDComposite. Got rid of a lot of duplicate code that way.


DanielGibson:
As far as I can tell, it's done and works (I tested with the USBHIDJoystick example which uses HIDUniversal).
I also tested HIDComposite with my own (so far unreleased) code for a USB-to-USB Keyboardadapter, which also still works.

Lauszus:
@DanielGibson thanks. I hope to do some testing this weekend.

Lauszus:
Works great. Thank you for the PR.

Issue #586, opened 2021-01-17
Bidirectional MIDI not working.


jhsa:
Hi, I am using an arduino Pro mini Atmega168 3.3V, 8Mhz. ALso, the Mini USB Host Shield.
If I flash the USB_Midi_coverter code, all works and I have a signal on the Arduino TX pin when I send a midi command via USB.

If I flash the Bidirectional_Converter code, I do not get a signal on the Arduino TX pin, and the Arduino LED just blinks repeatedly. A short puls followed by a longer pulse.
Any help would be very much appreciated. Thanks

YuuichiAkagawa:
It works on my arduino Pro mini Atmega328 3.3V. (with Arduino MIDI Library v5.0.2)

The bidirectional_converter depends on the behavior of the Arduino MIDI Library.
Does Arduino MIDI Library work on your board?

jhsa:
Hi, thanks for your reply.
Well, I have also the v.5.0.2 installed.
What connections need to be made between the shield and the arduino. Only The SPI, SS and RST? Do i also need INT? 
I don't have the boards on top of each other.
Thanks a lot for your time.

jhsa:
ok, Changed the board to another pro mini with m328 running at 16Mhz but powered by 3.3V, and it seems I have data on my scope. need to find out if the data is good of course :) Need also to try the other direction. Is it ok that we keep this issue open a bit longer? It might help others. There isn't much info about this bidirectional feature :( 
Thank you for your patience.

jhsa:
Ok, if I  send midi data into the host, I seem to have good data on the Arduino TX pin.
But if I send MIDI into the Arduino RX pin, I have bad data on the host output. 
I have a good working midi to USB converter cable connected to the host shield.
First picture shows the correct MIDI data for Program Change 17 on Channel 2.
The second picture shows bad data. it seems that there is data missing at the beginning of the data stream.
So, the code seems not to be processing the midi message correctly.

![Midi Output Correct](https://user-images.githubusercontent.com/5117455/104867356-f2ba3280-5940-11eb-98c4-52529045f0ed.jpg)

![Midi Output Wrong](https://user-images.githubusercontent.com/5117455/104867327-e3d38000-5940-11eb-84e1-a390e42f9fae.jpg)

YuuichiAkagawa:
> Only The SPI, SS and RST? Do i also need INT?

Please connect INT.

My board is works.

[...]

PR #596, opened 2021-01-28
Fix for handling MIDI channels in the bidirectional converter example

YuuichiAkagawa:
Fix for handling MIDI channels in the bidirectional converter example (#586)

### DIFF
--- a/examples/USBH_MIDI/bidirectional_converter/bidirectional_converter.ino
+++ b/examples/USBH_MIDI/bidirectional_converter/bidirectional_converter.ino
@@ -44,7 +44,7 @@ MIDI_CREATE_DEFAULT_INSTANCE();
 //////////////////////////
 
 USB Usb;
-USBHub  Hub1(&Usb);
+//USBHub  Hub1(&Usb);
 USBH_MIDI Midi(&Usb);
 
 void MIDI_poll();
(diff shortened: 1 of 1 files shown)


examples/USBH_MIDI/bidirectional_converter/bidirectional_converter.ino
```diff
@@ -87,6 +88,9 @@ void loop()
           //SysEx is handled by event.
           break;
         default :
+          if( msg[0] < 0xf0 ){
+            msg[0] |= MIDI.getChannel() - 1;
+          }
```
Lauszus, line 93: Could you add a comment explaining why this code is needed, so people less familiar with MIDI (including me) would be able to understand it?

Lauszus (changes requested):

examples/USBH_MIDI/bidirectional_converter/bidirectional_converter.ino
YuuichiAkagawa: - msg [0] is the MIDI status byte.
- Lower 4 bits of the status byte is the channel number.
- Status byte 0xFn is a system message.
- System messages do not have a channel number.

I forgot to set the channel number.
The getChannel() function in the Arduino MIDI library returns 1-16 as channel number.
https://fortyseveneffects.github.io/arduino_midi_library/a00033.html#ga0d2a0037fd9e4b06cafe7b2388451767
However, the channel number of raw MIDI messages is zero origin.

Lauszus: Okay. Can you add that comment to the example? Then I'll merge it :)

Lauszus (approved):

Lauszus:
Thanks for the PR. It's merged into master now.

PR #613, opened 2021-03-16
Add AMBX support

AranVink:
This PR is for adding support for the Philips AMBX lights. They are RGB light controlled by USB, using only a few simple output commands.

More information on the product:
[Review on Guru3D](https://www.guru3d.com/articles-pages/philips-ambx-gamers-kit-review,1.html)

I would totally understand if this is a niche product and not worth merging :)

P.s. The PS3USB was used as example to write this. It tried following the existing practices as much as possible.


examples/ambx/AMBX.ino
```diff
@@ -0,0 +1,55 @@
...
+
+    //Example using single light:
+    //AMBX.setLight(Wallwasher_center, White);
+
+    state++;
+    if (state > 4)
+      state = 0;
+    delay(1000);
+  }
+  delay(10);
```
Lauszus, line 54: What is the reason you have these delays? You should not block the main loop, as you need to call `Usb.Task` at a regular interval usually around every 1 ms.

Lauszus:
Thanks a lot for the PR. Additional drivers are always appreciated! I've left one comment on the code. Can you please take a look at it?

Also can you add a short section in the README.md as well?

examples/ambx/AMBX.ino
xxxajk: Yeah, those delays make no sense at all...

AranVink:
Updated the code based on your comments. Let me know what you think :)

Lauszus:
Thanks for the PR. Sorry for taking so long to merge it.

PR #614, opened 2021-03-22
Added support for Adafruit NRF52840 Feather Express 

josephduchesne:
Added definitions for Adafruit NRF52840 Feather Express based on the existing RBL_NRF51822 variant.
I left portions that apply to the whole NRF52 series under generic #ifdef sections to make adding other NRF52 series boards easier in the future. They should just require their own MAKE_PIN section and `typedef SPi` with board specific SPI pin configuration.

I was hoping to find a clean way to support generic NRF boards, but doing some sort of #define based loop seems like it would just make things hard to follow versus the straight forward but redundant MAKE_PIN section here.

The PS5 controller and USB_desc examples have been tested and seem to work fine with a knock off Duinofun USBmini 2.0 shield (after accounting for the mislabeled pins on this hardware, and cutting the USB VDD trace to wire it to 5V).

### DIFF
--- a/SPP.h
+++ b/SPP.h
@@ -114,7 +114,7 @@ public:
          */
         size_t write(const uint8_t* data, size_t size);
         /** Pull in write(const char *str) from Print */
-#if !defined(RBL_NRF51822)
+#if !defined(RBL_NRF51822) && !defined(NRF52_SERIES)
         using Print::write;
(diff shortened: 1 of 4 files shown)


Lauszus:
Thanks for the PR. I've just merged it :)

Issue #607, opened 2021-03-06
How to recover from OUT-NAK?


YuuichiAkagawa:
SNDFIFO is not cleared when USB::OutTransfer() exits when it reaches nak_limit.
Probably, when the next data is sent in this state, SNDFIFO becomes a double buffer operation.
https://youtu.be/ZxuxnGrov9U

![NAK-retransmission](https://user-images.githubusercontent.com/926923/110213745-af158b00-7ee4-11eb-9844-05df04f2735e.png)

When HRSLT returns non-zero, writing 0 to SNDBC seems to clear the SNDFIFO.
https://github.com/felis/USB_Host_Shield_2.0/compare/master...YuuichiAkagawa:pr_outnak_fix
https://github.com/YuuichiAkagawa/USB_Host_Shield_2.0/blob/c56eee9f163b25c2e4be76bd70f44bbf3c17cb8f/Usb.cpp#L390-L396

https://youtu.be/ZxuxnGrov9U?t=29
![NAK-discard](https://user-images.githubusercontent.com/926923/110213766-c785a580-7ee4-11eb-91c1-4569f71a16b7.png)

Lauszus:
Good find. Can you send a PR with the fix?

YuuichiAkagawa:
Yes I will. 
However, I have only tested MIDI driver + bulk transfer. Can anyone test in other cases?

Lauszus:
Fixed in #617

PR #617, opened 2021-03-26
Recover from OUT-NAK

YuuichiAkagawa:
SNDFIFO is not cleared when USB::OutTransfer() does not complete successfully.
See also #607 

It seems that we can recover with SNDBC=0 according to the following part of [APPLICATION NOTE 4000](https://www.maximintegrated.com/en/design/technical-documents/app-notes/4/4000.html).
> Programming Single-Buffered Transfers
> -- snip --
>    2. If NAK, go to the next step.
>  6. Reload the first FIFO byte, and relaunch the transfer:
>    1. Write SNDBC = 0, a dummy value to switch the FIFO containing the OUT data back under microcontroller control.

### DIFF
--- a/Usb.cpp
+++ b/Usb.cpp
@@ -388,6 +388,11 @@ uint8_t USB::OutTransfer(EpInfo *pep, uint16_t nak_limit, uint16_t nbytes, uint8
                 data_p += bytes_tosend;
         }//while( bytes_left...
 breakout:
+        /* If rcode(=rHRSL) is non-zero, untransmitted data remains in the SNDFIFO. */
+        if(rcode != 0) {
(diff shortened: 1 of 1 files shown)


Lauszus:
@YuuichiAkagawa thanks for the PR! I'm currently on vacation, but will test it with a few different controllers when I'm back.

xxxajk:
@YuuichiAkagawa  Nice find! I'll have to check this out with UHS3, since it's possible that it could solve some of the issues I am seeing in UHS3. Right now I'm way too busy to do any testing at all, but I will eventually try it out.

Lauszus:
I finally got around to merge this PR. Thanks again for finding and fixing the issue!

Issue #611, opened 2021-03-13
Naming conflicts with shared control identifiers (ButtonEnum)


dmadison:
Within the `ButtonEnum` used to list controls ([`controllerEnums.h`](https://github.com/felis/USB_Host_Shield_2.0/blob/28a75dea6b615b517205932e243c583fed85b5eb/controllerEnums.h)) the buttons are not unique. This creates a problem where two buttons with separate names but the same ID cannot be used on the same controller.

This occurs on the new Xbox Core controller used in the Series S / Series X, which has both a `VIEW` button (`4`, used with Xbox One) and a `SHARE` button (also `4`, used with PS4). Because they share the same value the `getButtonPress` function cannot distinguish between them.

To me it seems that the best solution would be to remove the values in the enum definition so that each control surface ID is unique, then each controller class could handle the parsing on a case-by-case basis. This would also allow you to do controller-specific transformations of inputs without affecting the other controller classes. The downside is that while this could maintain compatibility with programs that use the controls by name, it would unfortunately break compatibility with programs that use the enum values by number (if there are any).

I can look into making that change myself, but I'm unsure how much of a change to the public API you're comfortable with.

Lauszus:
I do not believe anybody ues the enums by number, so I'm fine with breaking up the enums.

A pull request with the changes would be awesome :)

PR #618, opened 2021-03-27
Make ButtonEnum values unique

dmadison:
Removes the number values associated with each `ButtonEnum` key so that all button identifiers are unique. This fixes issue #611 and potential future issues with the same problem.

Everywhere that referenced one of the enum values directly has been replaced with the `constexpr` function `ButtonIndex` which returns the previously associated value. Each controller now has its own `getButtonIndex` function which takes a `ButtonEnum` value and returns the relevant index value for that key, or `-1` if the key (or previously equivalent aliases) are not present on that controller.

This adds some overhead due to the function calls and also slightly modifies the public interface, because the numbered equivalents for the enums can no longer be used. All previously associated key aliases such as `SELECT` and `BACK` should still be maintained and work as they did before. This also adds bounds checking to all functions taking `ButtonEnum` as a parameter, so control indices out of range for a given controller now return '0' instead of attempting to read out of bounds memory.

I've tested this with the following controllers:
* Xbox Original (`XBOXOLD`)
* Xbox 360 Wired (`XBOXUSB`)
* Xbox 360 Wireless (`XBOXRECV`)
* Xbox Core Controller (`XBOXONE`)
* PS3 Wired (`PS3USB`)
* PS4 Wired (`PS4USB`)

Let me know if you see anything out of sorts.

### DIFF
--- a/PS3BT.cpp
+++ b/PS3BT.cpp
@@ -47,12 +47,12 @@ BluetoothService(p) // Pointer to USB class instance - mandatory
 }
 
 bool PS3BT::getButtonPress(ButtonEnum b) {
-        const int8_t index = getPS3ButtonIndex(b); if (index < 0) return 0;
+        const int8_t index = getButtonIndexPS3(b); if (index < 0) return 0;
         return (ButtonState & pgm_read_dword(&PS3_BUTTONS[index]));
 }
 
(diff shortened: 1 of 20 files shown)


Lauszus:
Thanks a lot for taking the time to make the PR. I've now merged it :)

PR #620, opened 2021-03-31
Implement InterfaceXtract() in confdescparser.h

YuuichiAkagawa:
I want use IntefaceXtract() function.

I think IntefaceXtract() needs to return a bool value, what do you think?

### DIFF
--- a/confdescparser.h
+++ b/confdescparser.h
@@ -30,7 +30,9 @@ e-mail   :  support@circuitsathome.com
 class UsbConfigXtracter {
 public:
         //virtual void ConfigXtract(const USB_CONFIGURATION_DESCRIPTOR *conf) = 0;
-        //virtual void InterfaceXtract(uint8_t conf, const USB_INTERFACE_DESCRIPTOR *iface) = 0;
+        virtual bool InterfaceXtract(uint8_t conf __attribute__((unused)), const USB_INTERFACE_DESCRIPTOR *iface __attribute__((unused))){
+                return true;
+        };
 
         virtual void EndpointXtract(uint8_t conf __attribute__((unused)), uint8_t iface __attribute__((unused)), uint8_t alt __attribute__((unused)), uint8_t proto __attribute__((unused)), const USB_ENDPOINT_DESCRIPTOR *ep __attribute__((unused))) {
         };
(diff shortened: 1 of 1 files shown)


xxxajk:
@YuuichiAkagawa welcome to the fun of dealing with UHS2 :-) 

Granted that UHS2 is designed to be as compact as possible, it may be best to have a stub method, which the end-user can fill in with any particulars. This lessens the loads on the developer, lowers the amount of code within. Alternetly, perhaps the correct thing to do is like is done in UHS3, where the sketch can use the midi class as a base-class, and then fork from there.

In any case, doing this sort of thing with limited resources it quite painful. USB is a pretty heavy lift for a 16MHz 8bit MCU, but that's sort-of what makes the challenge  fun. Perhaps it may be best to break away from a "one driver for all", such as how the game controllers are done? You will have to see what fits best, uses the least amount of RAM.

YuuichiAkagawa:
What I want to do is get the bInterfaceClass and bInterfaceSubClass in EndpointXtract() without affecting the existing driver.
Please let me know if there is a good way.

xxxajk:
Have you considered using the driver class as a base class, and then extending it? Big question is, ofcorse is UHS2 can actually do that. Even if it can, this will still consume RAM, but only for that particular sketch.

This is where UHS3 is actually a win, because it is capable of doing that. The downside is that UHS3 follows the USB specifications better, and because of that, needs to use more RAM. However, you can subclass right in the sketch, and totally alter the base class to your heart's content, and get exactly what you want, without modification to the library source at all.

Professionally, I only use one very specific device under UHS2, and tailor the library to only work with that device, and nothing else, and that's when I'm not using UHS3. When I use UHS3, I tend to use ARM or MIPS based MCU that has at least 64K RAM. Usually Freescale/NXP or Microchip devices.

USB can be a serious memory hog when attempting to support more than one device at a time, even when using UHS2, because it has to track a lot of information. 

The only other thing I could think of doing, which isn't cost effective whatsoever, would be to dedicate one small MCU just for very terse USB operations, such as detecting a connection, and assigning an address, and nothing else. Then use a second MCU to track what is going on, more-or-less as you would do using raw USB URBS. I don't know why you would want to do something like that though, when a 160MHz ARM is available for about $3US, and if you had to have USB within it (thus eliminating the MAX chip) for about $5US. And with that you usually get 1MB of flash, and 64k to 128k of RAM, sometimes more.

Sure, a tiny ATmega is fine for a very simple project that doesn't require a lot of speed or decisions, but once you start getting into something more complex, it's better to move to an MCU that fits the project.

That said, it's really up to you, and it fully depends on what the goal is.

YuuichiAkagawa:
withdraw

Issue #623, opened 2021-04-12
Xbox One S BT Controller does not seem to be connect if already paired


MasterJubei:
Hi everyone,
Using version 1.5.0
I am using an Arduino Uno for this test. I can do the initial pair the Xbox One S BT controller and it works just fine.  

However I cannot get the controller to connect without pairing. The controller is already paired, so I change the line XBOXONESBT Xbox(&Btd, PAIR); to XBOXONESBT Xbox(&Btd); but the controller can't connect.

In other words, the controller only works if I also pair it. It cannot reconnect without the controller and UHS in pair mode.

The PS4 controller example works just fine without needing to be paired each time.

When I set debugging on, this is the output of XBOXONESBT Xbox(&Btd, PAIR);

```
Xbox One S Bluetooth Library Started
Bluetooth Dongle Initialized
HCI Reset complete
Write class of device
Local Bluetooth Address: 5C:F3:70:A1:E5:EF
Dongle supports secure simple pairing (controller support)
Simple pairing was enabled
Set event mask completed
Please enable discovery of your device
Gamepad found
HID device found
Now just create the instance like so:
BTHID bthid(&Btd);
And then press any button on the device
Remote Name: Xbox Wireless Controller
Connecting to HID device
Connected to HID device
Received Key Request
Received IO Capability Request
SDP Incoming Connection Request
SDP Successfully Configured
Disconnect Request: SDP Channel
Disconnected SDP Channel
User confirmation Request
SDP Incoming Connection Request
SDP Successfully Configured
Disconnect Request: SDP Channel
Disconnected SDP Channel
SDP Incoming Connection Request
SDP Successfully Configured
Disconnect Request: SDP Channel
Disconnected SDP Channel
Pairing successful with HID device
Send HID Control Connection Request
Send HID Control Config Request
Set protocol mode: 00
Send HID Interrupt Connection Request
Send HID Interrupt Config Request
HID Channels Established
Wait For Incoming Connection Request
HCI Disconnected from Device
Wait For Incoming Connection Request
Gamepad is connecting
Incoming Connection Request
Remote Name: Xbox Wireless Controller
Connected to Device: C8:3F:26:46:19:47
Received Key Request
Wait For Incoming Connection Request
HCI Disconnected from Device
Wait For Incoming Connection Request
```

If I just have XBOXONESBT Xbox(&Btd);

```
Xbox One S Bluetooth Library Started
Bluetooth Dongle Initialized
HCI Reset complete
[...]

PR #624, opened 2021-04-16
Xbox One S BT Controller patch so it can reconnect

MasterJubei:
Hi, I noticed with the Xbox One S BT controller that is unable to reconnect. It only works if it is also being paired. This is because it does not save the link key.
https://github.com/felis/USB_Host_Shield_2.0/issues/623

Most of the hard work was done by @HisashiKato

This patch saves the bluetooth address and link key to EEPROM. I don't think there is a way around this.
The patch also checks if an Xbox One S BT controller is connecting so it handles it appropriately and does not break other controllers. 

I tested with a PS4 controller after making these changes, it can pair and can reconnect just fine.

This will break support for the STM32, although that can be fixed pretty easily by including an arduino EEPROM library. I am not sure about other microcontrollers.

### DIFF
--- a/BTD.cpp
+++ b/BTD.cpp
@@ -639,11 +639,24 @@ void BTD::HCI_event_task() {
                                 }
                                 break;
 
-                        case EV_LINK_KEY_REQUEST:
+                        case EV_LINK_KEY_REQUEST: //For the Xbox One S Controller we can't use the negative reply, other controllers use negative_reply
 #ifdef DEBUG_USB_HOST
                                 Notify(PSTR("\r\nReceived Key Request"), 0x80);
 #endif
-                                hci_link_key_request_negative_reply();
+                                if ((!pairWithHIDDevice || incomingHIDDevice) && incomingXboxOneS) {
+                                        for(uint8_t i = 0; i < 16; i ++) {
+                                                link_key[i] = EEPROM.read(i+6);
+                                        }
+                                        hci_link_key_request_reply();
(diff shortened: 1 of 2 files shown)


Lauszus:
The reason why I didn't do this was because not all boards supports the Arduino EEPROM API. For instance the ESP8622 requires you to call `EEPROM.begin` to pre-allocate some flash to be used for EEPROM and you also need to call `EEPROM.commit` to actually store the content.

I think it would be better to have a callback and then deal with writing and reading from the EEPROM in the `ino` file. I'll make a new PR based on this.

Lauszus:
Also I would rather not hardcode the EEPROM address deep down in the library, as it will be quite confusing for people if their EEPROM data is overridden.

Anyway I'll base my code on yours, but with a callback instead :)

MasterJubei:
@Lauszus 
Yup makes sense. The callback is a good idea, that would solve that. I was thinking along the lines of checking if the board is an arduino uno, that you would call that. Your idea is better.

As for hadcoding the location to write to...
_At least their program data would go to a good cause._  /s

There is one more issue with the Xbox One S BT controller. The controller data has significantly changed with a firmware update, so a lot of the buttons are mapped incorrectly.
https://github.com/felis/USB_Host_Shield_2.0/issues/627#issuecomment-826255632

You can access all the data with the least amount of effort by doing what I did instead, but the way Hisashi did it, is the correct way. I will see if I can eventually make a PR request doing it the proper way.

PR #640, opened 2021-05-28
Added support for Adafruit NRF52832 Feather

ar664:
Used this pr as a reference:
https://github.com/felis/USB_Host_Shield_2.0/pull/614

I successfully flashed it and it went through USB.Init without error. 
Will report back when I have tested it with my original xbox controller. If anyone would like to test it in the mean time, feel free to do so.

### DIFF
--- a/avrpins.h
+++ b/avrpins.h
@@ -1295,6 +1295,65 @@ MAKE_PIN(P19, GPIOC, GPIO_PIN_0); // A5
 #undef MAKE_PIN
 
 
+#elif defined(ARDUINO_NRF52832_FEATHER)
+
+#define MAKE_PIN(className, pin) \
+class className { \
+public: \
+    static void Set() { \
+        nrf_gpio_pin_set(pin); \
+    } \
+    static void Clear() { \
+        nrf_gpio_pin_clear(pin); \
+    } \
+    static void SetDirRead() { \
+        nrf_gpio_cfg_input(pin, NRF_GPIO_PIN_NOPULL); \
+    } \
+    static void SetDirWrite() { \
+        nrf_gpio_cfg_output(pin); \
+    } \
+    static uint8_t IsSet() { \
+        return (uint8_t)nrf_gpio_pin_read(pin); \
+    } \
+};
+
+// Based on variants/feather_nrf52832/variant.cpp
+// g_ADigitalPinMap could be used directly, but it would be slower
+MAKE_PIN(P0, (0));
+MAKE_PIN(P1, (1));
+MAKE_PIN(P2, (2));
+MAKE_PIN(P3, (3));
+MAKE_PIN(P4, (4));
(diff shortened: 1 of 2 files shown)


Lauszus:
Looks good. Can you also add it to the list of supported boards: https://github.com/felis/USB_Host_Shield_2.0#boards and the CI: https://github.com/felis/USB_Host_Shield_2.0/blob/master/.github/workflows/main.yml#L48?

ar664:
Would the CI be like this?
```
pio ci --lib="." --board=adafruit_feather_nrf52832 --project-option="build_flags=-Wno-sign-compare -Wno-unused-function -Wno-unused-variable"
```
How do I test this?

Lauszus:
Simply add it to the other NRF52 board i.e:

```yml
pio ci --lib="." --board=adafruit_feather_nrf52840 --board=adafruit_feather_nrf52832 --project-option="build_flags=-Wno-sign-compare -Wno-unused-function -Wno-unused-variable"
```

Then commit and push it and the CI will trigger and build it automatically.

ar664:
It working was a false positive. I ran the board_qc sketch and I get this. 
```
Reading REVISION register... Die revision invalid. Value returned: FF
Unrecoverable error - test halted!!
0x55 pattern is transmitted via SPI
```
I have the duinofun mini board and have the SPI connected correctly after checking the pins with the MAX3421E datasheet. Do I need to connect the Reset pin in the MAX3421E

ar664:
Do I need to modify my setup with any pull down resistors? 

- Serial is working
- SPI is correctly wired, given MAX3421E spec sheet
- Nrf52832 spec says it's able to output 3.3v 500ma, so I assume the host is able to be powered
- The SS Pin measured 3.3V
- I don't have oscilloscope, but I measure a peak of .4V from the SPI signals.

Don't mind the mess, but here is a visual of the current setup. (If you need a diagram lmk).

Nrf52832Setup

In addition to the before example 00 is also returned by both sometimes with board_qc.
```
Reading REVISION register... Die revision invalid. Value returned: 00
Unrecoverable error - test halted!!
0x55 pattern is transmitted via SPI
```

Both of them come up with the same result with XBOXONE example:
```
OSC did not start
```
or
```
XBOX USB Library Started
BSP Library : 0.22.1
Bootloader  : s132 6.1.1
Serial No   : FC7C7CD0DB9ECC34
```
but no input is detected and no white light indicator on the xbox controller.

The xbox controller does vibrate indicating that power was received.
I'm using a powered usb hub to power the whole thing with an external power supply, should be good up to 5v 2000ma. I've ran usb hdd's off of it.

xxxajk:
If you still get unreliable operation, there's two thins that can cause this that many people do not realize.
1: Keep SPI lines as short as possible
2: If you can't use shorter wires, lower the SPI speed.
3: A combination of 1 and 2.

Issue #646, opened 2021-06-24
hidboot:  Poll() buffer overrun


tmk:
Value from MaxPacketSize of Endpoint Descriptor is not checked before using, which can be upto 64 legally in full-speed, or more with malicious or buggy implementation.

Does this patch make sense?

```
diff --git a/hidboot.h b/hidboot.h
index 4076400..9cb144d 100644
--- a/hidboot.h
+++ b/hidboot.h
@@ -582,7 +582,7 @@ uint8_t HIDBoot<BOOT_PROTOCOL>::Poll() {
 
                 // To-do: optimize manually, using the for loop only if needed.
                 for(int i = 0; i < epMUL(BOOT_PROTOCOL); i++) {
-                        const uint16_t const_buff_len = 16;
+                        const uint16_t const_buff_len = 64;
                         uint8_t buf[const_buff_len];
 
                         USBTRACE3("(hidboot.h) i=", i, 0x81);
@@ -590,6 +590,9 @@ uint8_t HIDBoot<BOOT_PROTOCOL>::Poll() {
                         USBTRACE3("(hidboot.h) epInfo[epInterruptInIndex + i].maxPktSize=", epInfo[epInterruptInIndex + i].maxPktSize, 0x81);
                         uint16_t read = (uint16_t)epInfo[epInterruptInIndex + i].maxPktSize;
 
+                        if (read > const_buff_len)
+                            read = const_buff_len;
+
                         rcode = pUsb->inTransfer(bAddress, epInfo[epInterruptInIndex + i].epAddr, &read, buf);
                         // SOME buggy dongles report extra keys (like sleep) using a 2 byte packet on the wrong endpoint.
                         // Since keyboard and mice must report at least 3 bytes, we ignore the extra data.
```

Lauszus:
Good catch. Wouldn't it be better just to set the size of `buf` dynamically i.e:

```C
uint8_t buf[read];
```

tmk:
Nice idea. We can save stack space with that, instead of declaring 64 bytes always.

I tested this change with examples/HID/USBHIDBootKbd and confirmed that it works as expected.

@Lauszus 
Could you make a commit for this small change? Or is that better I make a PR?

```diff --git a/hidboot.h b/hidboot.h
index 4076400..4fe83ff 100644
--- a/hidboot.h
+++ b/hidboot.h
[...]

PR #647, opened 2021-06-26
hidboot: Fix buffer size and check maxPacketSize #646

tmk:
This fixes buffer handling in `Poll()` of hidboot.h. #646

`Poll()` can cause buffer overrun in `inTransfer()` becase it uses `maxPktSize` as buffer size without checking.
https://github.com/felis/USB_Host_Shield_2.0/blob/08121e1701908006cd646ad51b8c93cf7fe4ce7a/hidboot.h#L585-L593


Lauszus:
Thanks for the bug fix. I've now merged it :)

PR #650, opened 2021-07-01
Add Xbox Trigger Enums

dmadison:
This adds `LB/RB` and `LT/RT` enum values for the Xbox controllers, in place of the original `L1/R1` and `L2/R2` keys. These are how Microsoft refers to the controls, and what is molded into the plastic on the Xbox 360 controllers. I've also updated the Xbox examples to swap out the keys and related strings.

While I was modifying the Original Xbox code, I also changed a duplicate reference to the VID/PID check and added support for the "Logitech Cordless Precision" controller.

This change uses the same button index value as before, so it should be backwards compatible with previous versions. The old key values can also still be used without issue.

### DIFF
--- a/XBOXOLD.cpp
+++ b/XBOXOLD.cpp
@@ -118,7 +118,7 @@ uint8_t XBOXOLD::Init(uint8_t parent, uint8_t port, bool lowspeed) {
         VID = udd->idVendor;
         PID = udd->idProduct;
 
-        if((VID != XBOX_VID && VID != MADCATZ_VID && VID != JOYTECH_VID) || (PID != XBOX_OLD_PID1 && PID != XBOX_OLD_PID2 && PID != XBOX_OLD_PID3 && PID != XBOX_OLD_PID4)) // Check if VID and PID match
+        if(!VIDPIDOK(VID, PID)) // Check if VID and PID match
                 goto FailUnknownDevice;
 
         // Allocate new address according to device class
--- a/XBOXOLD.cpp
+++ b/XBOXOLD.cpp
@@ -293,6 +293,19 @@ void XBOXOLD::printReport(uint16_t length __attribute__((unused))) { //Uncomment
 }
 
 int8_t XBOXOLD::getAnalogIndex(ButtonEnum b) {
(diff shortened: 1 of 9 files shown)


XBOXOLD.cpp
```diff
@@ -293,6 +293,19 @@ void XBOXOLD::printReport(uint16_t length __attribute__((unused))) { //Uncomment
...
 
 int8_t XBOXOLD::getAnalogIndex(ButtonEnum b) {
+        // For legacy reasons these mapping indices not match up,
+        // as the original code uses L1/R1 for the triggers and
+        // L2/R2 for the white/black buttons. To fix these new enums
+        // we have to transpose the keys before passing them through
+        // the button index function
+        switch (b) {
+        case(LT): b = L1; break;  // normally L2
+        case(RT): b = R1; break;  // normally R2
```
Lauszus, line 303: Shouldn't this be `L2` and `R2` or have I misunderstood it?

Lauszus:
Thanks for the pull request! I've left one comment on your code.

Lauszus:
Btw don't worry about the CI failing. I'll fix that after merging the pull request.

XBOXOLD.cpp
dmadison: It's correct, it just looks a little odd. My comment probably could have explained it better.

The library's implementation of the original Xbox controller is a little wonky because of the shared enum IDs. The controller only has one set of triggers so those were programmed as `L1/R1`, and the white/black buttons that became `LB/RB` on the 360 controller were programmed with the same IDs as `L2/R2`. That creates a conflict because the other Xbox controllers use the reverse - `L2/R2` for the triggers and `L1/R1` for the option buttons (bumpers).

Without this little filter that means `LT/RT` returns the white/black buttons, and `LB/RB` return the triggers. The filter swaps the enum values around before it runs through the ID filter, so whether `L1` is passed (previous versions) or `LT` is passed (current versions) the trigger is still outputted.

Lauszus:
I've now merged your PR. Again thanks for submitting it :)

Luxruy (commented):
#796

PR #653, opened 2021-07-25
Improve plugin detection in busprobe()

tmk:
I like to share this `busprobe()` patch and need code review and your thought on it.

Seems like MAX3421e `CONDETIRQ`(Connect/Disconnect Interrupt) doesn't arise for every plugin event and the chip fails to detect it sometimes.  (Meanwhile, I never seen the chip failed in detection of unpluging.)

For confirmaion I tried to catch INT pin change and read bus state immedately using `ISR()`, but it still failed to detect device insertion event occasionally. So I believe that we cannot rely solely on the `CONDETIRQ` interrupt for plugin detection.

This patch checks bus state at every `Task()` call when `vbusState` is  `SE0`(no device) regardless of INT pin.
There are some descriptions of the fix here also. https://github.com/tmk/USB_Host_Shield_2.0/issues/9

These issues are related probably. felis#425 felis#452

### DIFF
--- a/usbhost.h
+++ b/usbhost.h
@@ -503,9 +503,28 @@ int8_t MAX3421e< SPI_SS, INTR >::Init(int mseconds) {
 /* probe bus to determine device presence and speed and switch host to this speed */
 template< typename SPI_SS, typename INTR >
 void MAX3421e< SPI_SS, INTR >::busprobe() {
+        static uint8_t prev_bus = -1;
         uint8_t bus_sample;
+
+        if (regRd(rHIRQ) & bmCONDETIRQ) {
+            // device plug/unplug is detected
+            // clear CONDETIRQ
+            regWr(rHIRQ, bmCONDETIRQ);
+        } else {
+            // check current bus state when no device(SE0),
+            // otherwise transient state will be read due to bus activity.
+            if (vbusState != SE0) return;
+
+            // read current bus state
+            regWr(rHCTL, bmSAMPLEBUS);
+        }
+
         bus_sample = regRd(rHRSL); //Get J,K status
         bus_sample &= (bmJSTATUS | bmKSTATUS); //zero the rest of the byte
+
+        if (prev_bus == bus_sample) { return; }
+        prev_bus = bus_sample;
+
         switch(bus_sample) { //start full-speed or low-speed host
                 case( bmJSTATUS):
                         if((regRd(rMODE) & bmLOWSPEED) == 0) {
@@ -538,6 +557,7 @@ void MAX3421e< SPI_SS, INTR >::busprobe() {
 /* MAX3421 state change task and interrupt handler */
 template< typename SPI_SS, typename INTR >
 uint8_t MAX3421e< SPI_SS, INTR >::Task(void) {
+        /*
         uint8_t rcode = 0;
         uint8_t pinvalue;
         //USB_HOST_SERIAL.print("Vbus state: ");
@@ -553,6 +573,9 @@ uint8_t MAX3421e< SPI_SS, INTR >::Task(void) {
         //    }
         //    usbSM();                                //USB state machine
         return ( rcode);
+        */
+        busprobe();
+        return 0;
 }
 
 template< typename SPI_SS, typename INTR >


usbhost.h
```diff
@@ -503,9 +503,28 @@ int8_t MAX3421e< SPI_SS, INTR >::Init(int mseconds) {
...
+            // device plug/unplug is detected
+            // clear CONDETIRQ
+            regWr(rHIRQ, bmCONDETIRQ);
+        } else {
+            // check current bus state when no device(SE0),
+            // otherwise transient state will be read due to bus activity.
+            if (vbusState != SE0) return;
+
+            // read current bus state
+            regWr(rHCTL, bmSAMPLEBUS);
```
Lauszus, line 519: Shouldn't you wait for the operation to finish like how it is done elsewhere in the code?

I.e:

```C++
while(!(regRd(rHCTL) & bmSAMPLEBUS)); //wait for sample operation to finish
```

Lauszus:
Sounds good to me, as the overhead is minimal.

usbhost.h
tmk: That code seems to use contrary logic according to MAX3421E reference?
From my tests the waiting code does nothing and is not needed in fact.

The chip reference(AN3785.pdf p.52) says:
> The CPU sets the SAMPLEBUS bit to instruct the SIE to sample the state of the D+ and D-
lines. The SIE clears the SAMPLEBUS bit when it completes the operation.

But the `SAMPLEBUS` bit shows strange behaviour there in fact.

I never seen 'S' with code below.

```while (!(regRd(rHCTL) & bmSAMPLEBUS)) { Serial.print("S "); };```

While this code is repeated forever for some reason.

```while (regRd(rHCTL) & bmSAMPLEBUS) { Serial.print("S "); }; ```

File: usbhost.h [hunk 2]
```diff
@@ -553,6 +573,9 @@ uint8_t MAX3421e< SPI_SS, INTR >::Task(void) {
         //    }
         //    usbSM();                                //USB state machine
         return ( rcode);
+        */
+        busprobe();
```
Lauszus, line 577: Maybe add a comment why the code above was commented out and replaced with just `busprobe`.

usbhost.h
Lauszus: I guess the SAMPLEBUS bit is cleared in between the two calls.

tmk: With some more tests the SAMPLEBUS bits doesn't seems to be cleared until bus state is changed but not always. This is clearly not what the reference describes, we can't use the bit to wait for sampling bus state anyway.

And simply asserting SAMPLEBUS everytime before reading J/KSTATUS is good practice seemingly. This is what MaximUSBLab10 code does.

My code works in my tests at least so far.

usbhost.h
tmk: Is this proper for the comment? If so I'll commit it.

```
// CONDETIRQ can miss detection of plugin and check bus every time anyway #653
```

Lauszus: What about something like this:

```C++
/* MAX3421 state change task and interrupt handler */
template< typename SPI_SS, typename INTR >
uint8_t MAX3421e< SPI_SS, INTR >::Task(void) {
#if 0
        uint8_t rcode = 0;
        uint8_t pinvalue;
        //USB_HOST_SERIAL.print("Vbus state: ");
        //USB_HOST_SERIAL.println( vbusState, HEX );
        pinvalue = INTR::IsSet(); //Read();
        //pinvalue = digitalRead( MAX_INT );
        if(pinvalue == 0) {
                rcode = IntHandler();
        }
        //    pinvalue = digitalRead( MAX_GPX );
        //    if( pinvalue == LOW ) {
        //        GpxHandler();
        //    }
        //    usbSM();                                //USB state machine
        return ( rcode);
#else
        // Seems like MAX3421e CONDETIRQ (connect/disconnect Interrupt) doesn't arise for every plugin event, 
        // so instead we check the bus state at every call to Task when no device is connected
        // See: https://github.com/felis/USB_Host_Shield_2.0/pull/653
        busprobe();
#endif
        return 0;
}
```

I think the above code makes it more clear what the old vs new code is as well.

tmk: It looks good!
Please feel free to add it as new commit or change existent commit before merging.

Thanks for your service on this library as always!

Issue #676, opened 2021-11-28
PS4 Controller unable to pair.


benjaminaarsen:
Hello, i have been trying to make this library work for a few hours now. But i don't seem to be able to get it working.
This is the output i get.

PS4 Bluetooth Library Started
Bluetooth Dongle Initialized
HCI Reset complete
Write class of device
Local Bluetooth Address: 00:1A:7D:DA:71:10
Please enable discovery of your device
Couldn't find HID device
Wait For Incoming Connection Request

The bluetooth dongle is flashing, and the host shield is working fine as well. Since the PS4USB works perfectly fine. It just somehow can't find the controller. When i use the dongle on my pc it is able to find the controller though... So communication with the dongle must be going wrong somewhere.

This is the dongle i used: https://www.aliexpress.com/item/33015867058.html?src=google&src=google&memo1=freelisting&albch=shopping&acnt=494-037-6276&slnk=&plac=&mtctp=&albbt=Google_7_shopping&albagn=888888&isSmbAutoCall=false&needSmbHouyi=false&albcp=6459980570&albag=76980386066&trgt=539263010115&crea=nl33015867058&netw=u&device=c&albpg=539263010115&albpd=nl33015867058&gclid=Cj0KCQiA7oyNBhDiARIsADtGRZYXVA8mmw_XzTdPg1HgpW-2Et6MjnLZF6As2fPoB3IW760sxFrsa4QaAj98EALw_wcB&gclsrc=aw.ds&aff_fcid=422fe2820d0f4c6cba418eace4f859a6-1638099471713-04895-UneMJZVf&aff_fsk=UneMJZVf&aff_platform=aaf&sk=UneMJZVf&aff_trace_key=422fe2820d0f4c6cba418eace4f859a6-1638099471713-04895-UneMJZVf&terminal_id=2f3d62296a4b48e4a0e29328d7ed07b4.

I would really appreciate if you could help me.

xxxajk:
If it requires firmware, then you are out of luck. UHS2 only works with "dumb" bluetooth dongles, and those may be hard to find.

benjaminaarsen:
How do i know whether the bluetooth dongle requires firmware or not? And do you perhaps know where to find the list of compatible bluetooth dongles? It was referenced on a few forums, but i can't seem to find it

Lauszus:
@xxxajk the dongle is recognized, so it does not require any firmware.
[...]

PR #677, opened 2021-12-02
Set inquiry mode to "Standard Inquiry Result event format" before doing a HCI inquiry

Lauszus:
Fixes: #676


Lauszus:
Superseded by #681

Issue #688, opened 2021-12-30
After using Midi:SendData() the library stops receiving messages on the second virtual interface


chrisp250:
Hi,
I'm using the library with a mini USB shield 2.0 and a Teensy 3.2. Also SPI4TEENSY3 is disabled.

I'm trying to communicate with a Novation Launchpad Mini MK3. This devices has two virtual interfaces for midi.
This is a summary of the code:
Midi:SendSysEx()
loop:
  Midi:RecvData()

This works fine and RecvData() received data on both ports.
However, if I instead do this:
Midi:SendData()
loop:
  Midi:RecvData()

I can no longer receive any events coming on the second virtual interface. The first virtual interface is fine and I can receive the notes coming from the pads.
I tested the messages on a PC with the same device and it works fine, so I'm confident this is not a device issue.
Also I need be able to send "Note On" messages to the device, so just replacing SendData with SendSysEx is not a usable workaround in my case.

Thank you in advance
Christian

Lauszus:
@YuuichiAkagawa can you take a look at this?

YuuichiAkagawa:
The two MIDI ports use the same endpoint and are not distinguished by the MIDI driver.
- CN=0 : LPMiniMK3 DAW
- CN=1 : LPMiniMK3 MIDI

If you can't get the value, it's likely that the device isn't sending.
In what mode are you operating? Is it in DAW mode? Or is it Programmer mode?

chrisp250:
Hi Yuuichi,
Thank you for your reply.

> The two MIDI ports use the same endpoint and are not distinguished by the MIDI driver.
> 
> * CN=0 : LPMiniMK3 DAW
> * CN=1 : LPMiniMK3 MIDI

Yes, that's correct.
 
> If you can't get the value, it's likely that the device isn't sending. In what mode are you operating? Is it in DAW mode? Or is it Programmer mode?

I'm confident the launchpad still sends messages because I tested it sending the same exact messages using a PC/python combination and it works fine. I have tested on both DAW and Live mode and the behaviour is the same. As soon as I do a _SendData()_ I stop getting messages on **CN=1 : LPMiniMK3 MIDI**.

I put together a sample/simplified code hoping it will help identifying the issue:

`
/*
 *******************************************************************************
 * USB-MIDI Launchpad Mini mk3 test
 *
 *******************************************************************************
 */

#define SPI4TEENSY3 0
#include <Arduino.h>
#include <usbh_midi.h>
[...]

PR #689, opened 2022-01-06
Fix for non-zero CN issue(#688)

YuuichiAkagawa:
If the CableNumber(CN) is non-zero, RecvData () responds with the wrong message size.
Changed the method to get MIDI message size in version 0.6.0, but the argument was wrong.

In the first byte of the MIDI Event packet, CN is the upper 4 bits and CIN is the lower 4 bits.
Only the lower 4 bits must be given to getMsgSizeFromCin().


Lauszus:
Thanks :)

PR #693, opened 2022-02-06
Add support for MightyCore

rharke:
This adds support for [MightyCore](https://github.com/MCUdude/MightyCore), an Arduino core for a number of larger 32 I/O AVR microcontrollers.

It has 3 selectable pin mapping configurations for compatibility with some other boards. This change should support all 3, though I've only actually tested the "standard" configuration.


Lauszus:
Thanks for the contribution! Could you also add the board to the CI?

rharke:
Done.

Lauszus:
Please skip the Wii and PS3SPP example like I have done for the Uno: https://github.com/felis/USB_Host_Shield_2.0/blob/786b7daae2ae7504d03946ef2fbf5ce3f88a1840/.github/workflows/main.yml#L30

As the examples does not fit on the ATmega32: https://github.com/felis/USB_Host_Shield_2.0/runs/5089727646?check_suite_focus=true#step:7:279

Lauszus:
I merged your changes in https://github.com/felis/USB_Host_Shield_2.0/pull/708

Thanks for your contribution!

PR #705, opened 2022-04-21
Update MIDI driver v1.0.0

YuuichiAkagawa:
MIDI Driver v1.0.0
* Add OnRelease() callback.
* The timing for enabling PollEnable has been changed to before the onInit() callback.
* Update vender specific code for Novation.

MIDI Driver celebrates its 10th anniversary. :tada:


Lauszus:
Thanks!

PR #722, opened 2022-06-03
Removed trailing comma in library.json

gabryelreyes:
Fixes Library Manager warning:
Library Manager: Could not parse manifest -> Expecting property name enclosed in double quotes: line 10 column 5 (char 332)


Lauszus:
Thanks!

PR #737, opened 2022-09-06
Add functionality to retrieve input source from MiniDSP 2x4HD

dennisfrett:
Additional functionality for the MiniDSP driver.

Next to volume and muted status, it can now also retrieve the current input source (Analog, Toslink or USB).

### DIFF
--- a/MiniDSP.cpp
+++ b/MiniDSP.cpp
@@ -24,6 +24,7 @@
 void MiniDSP::ParseHIDData(USBHID *hid __attribute__ ((unused)), bool is_rpt_id __attribute__ ((unused)), uint8_t len, uint8_t *buf) {
 
         constexpr uint8_t StatusInputCommand[] = {0x05, 0xFF, 0xDA};
+        constexpr uint8_t InputSourceInputCommand[] = {0x05, 0xFF, 0xD9};
 
         // Only care about valid data for the MiniDSP 2x4HD.
         if(HIDUniversal::VID != MINIDSP_VID || HIDUniversal::PID != MINIDSP_PID || len <= 4 || buf == nullptr)
@@ -52,6 +53,28 @@ void MiniDSP::ParseHIDData(USBHID *hid __attribute__ ((unused)), bool is_rpt_id 
                 if(pFuncOnMutedChange != nullptr && mutedChanged)
                         pFuncOnMutedChange(muted);
         }
+
+
+        // Check if this is an input source update.
+        // First byte is the length, we ignore that for now.
+        if(memcmp(buf + 1, InputSourceInputCommand, sizeof (InputSourceInputCommand)) == 0) {
+
+                // Parse data.
+                // Response is of format [ length ] [ 0x05 0xFF 0xD9 ] [ source ].
+                const auto newInputSource = buf[sizeof (InputSourceInputCommand) + 1];
+
+                // Ensure we only interpret valid inputs.
+                if(newInputSource >= 0x00 && newInputSource <= 0x02){
+                        const auto inputSourceChanged = newInputSource != (char) inputSource;
+
+                        // Update values.
+                        inputSource = (InputSource) newInputSource;
+
+                        // Call callbacks.
+                        if(pFuncOnInputSourceChange != nullptr && inputSourceChanged)
+                                pFuncOnInputSourceChange(inputSource);
+                }
+        }
 };
 
 uint8_t MiniDSP::OnInitSuccessful() {
@@ -59,8 +82,9 @@ uint8_t MiniDSP::OnInitSuccessful() {
         if(HIDUniversal::VID != MINIDSP_VID || HIDUniversal::PID != MINIDSP_PID)
                 return 0;
 
-        // Request current status so we can initialize the values.
+        // Request current information so we can initialize the values.
         RequestStatus();
+        RequestInputSource();
 
         if(pFuncOnInit != nullptr)
                 pFuncOnInit();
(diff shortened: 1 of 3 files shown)


examples/MiniDSP/MiniDSP.ino
```diff
@@ -25,6 +25,20 @@ void OnMutedChange(bool isMuted) {
   Serial.println("Muted status: " + String(isMuted ? "muted" : "unmuted"));
 }
 
+void OnInputSourceChange(MiniDSP::InputSource inputSource) {
+  String inputSourceStr;
```
Lauszus, line 29: Maybe initialize this to "Unknown"?

Lauszus:
Looking good. I've left one comment. Can you address this and then I'll merge it :)

dennisfrett:
> I added also the functionality to retrieve current config from MiniDSP 2x4HD.

I can check if I can add these later this week. @Lauszus I don't know if you'd rather merge this change in already, or if you'd like to merge everything together later?

MarkLido:
Better later. After the function for "send" is added!?

Lauszus:
Okay great. Just let me know when it's ready to get merged :)

dennisfrett:
I updated the code to also retrieve the current config + I fixed the callback for input source not being called (Might be handy for you, @MarkLido ).

Updating volume has not been done yet.

However, @Lauszus , I'm having an issue, in `OnInitSuccessful` I do:
```lang=cpp
        RequestStatus();
        RequestInputSource();
        RequestConfig();
```

Which seems to be sending commands, but not listening for a response,  so most of the time the response to the first command (request status), is just lost.

What would you suggest here?

I guess I'm using the library wrong, should I use `ctrlReq` instead? Or should I somehow busy wait for a response myself?

MarkLido:
Thanks Dennis. The callback for input source works now.

dennisfrett:
@MarkLido I've made a quick implementation for setting volume and muted status.

But I still think I might not be handling synchronous commands correctly, so any input from @Lauszus here would be awesome.

MiniDSP.cpp
```diff
@@ -109,3 +163,33 @@ void MiniDSP::RequestStatus() const {
...
+        uint8_t SetVolumeCommand[] = {0x42, (uint8_t)(-2*volumeDB)};
+
+        SendCommand(SetVolumeCommand, sizeof(SetVolumeCommand));
+}
+
+void MiniDSP::setMuted(bool muted) const {
+        uint8_t SetMutedommand[] = {0x17, muted ? (uint8_t)0x01 : (uint8_t)0x00};
+
+        SendCommand(SetMutedommand, sizeof(SetMutedommand));
+}
```
Lauszus, line 195: Missing new line.

MiniDSP.cpp
```diff
@@ -109,3 +163,33 @@ void MiniDSP::RequestStatus() const {
...
+
+void MiniDSP::setVolumeDB(float volumeDB) const {
+        // Only accept values between 0dB and -127.5dB.
+        // Don't do error handling.
+        if(volume > 0 || volume < -127.5){
+                return;
+        }
+
+        uint8_t SetVolumeCommand[] = {0x42, (uint8_t)(-2*volumeDB)};
```
Lauszus, line 186:
```suggestion
       uint8_t SetVolumeCommand[] = {0x42, (uint8_t)round(-2.0f * volumeDB)};
```

MiniDSP.cpp
```diff
@@ -109,3 +163,33 @@ void MiniDSP::RequestStatus() const {
...
+void MiniDSP::RequestConfig() const {
+        uint8_t RequestConfigCommand[] = {0x05, 0xFF, 0xD8, 0x01};
+
+        SendCommand(RequestConfigCommand, sizeof(RequestConfigCommand));
+}
+
+void MiniDSP::setVolumeDB(float volumeDB) const {
+        // Only accept values between 0dB and -127.5dB.
+        // Don't do error handling.
+        if(volume > 0 || volume < -127.5){
```
Lauszus, line 182:
```suggestion
        if(volume < 0 || volume > -127.5) {
```

MiniDSP.h
```diff
@@ -159,6 +219,18 @@ class MiniDSP : public HIDUniversal {
         void
         RequestStatus() const;
 
+        /**
+         * Send the "Request input source" command to the MiniDSP.
+         */
+        void
```
Lauszus, line 225:
```suggestion
        void RequestInputSource() const;
```

MiniDSP.h
```diff
@@ -159,6 +219,18 @@ class MiniDSP : public HIDUniversal {
...
+        RequestInputSource() const;
+
+        /**
+         * Send the "Request config" command to the MiniDSP.
+         */
+        void
```
Lauszus, line 231:
```suggestion
        void RequestConfig() const;
```

Lauszus:
> I updated the code to also retrieve the current config + I fixed the callback for input source not being called (Might be handy for you, @MarkLido ).
> 
> Updating volume has not been done yet.
> 
> However, @Lauszus , I'm having an issue, in `OnInitSuccessful` I do:
> 
> ```
>         RequestStatus();
>         RequestInputSource();
>         RequestConfig();
> ```
> 
> Which send commands, but don't listening for a response, so most of the time the response to the first command (request status), is just lost.
> 
> What would you suggest here?
> 
> I guess I'm using the library wrong, should I use `ctrlReq` instead? Or should I somehow busy wait for a response myself?

An workaround would be to just add `delay(1);` in between the calls.

dennisfrett:
@Lauszus Writing this driver without a way too easily debug USB pakets seems quite difficult, and my current implementation has some bugs (`delay`ing inbetween sending doesn't resolve the issue, setting volume only works the first time and not the second time, ...). To make my life easier I've purchased an Atmega32u4 in the hope I can set it up as a USB sniffer (https://github.com/matlo/serialusb). 

Once I figure out what's going wrong and fix it, I'll update the PR.

PR #756, opened 2022-12-12
Replace wrong trigger EffectType

einspunktnull:
this feature didn't work until replaced with correct enum value


Lauszus:
Thanks!

PR #779, opened 2023-07-14
Updated for Arduino Uno R4 maxima and wifi

mjs513:
Updated for latest Arduino R4 boards (maxima/wifi).

Tested BT with PS4 and XBox, and wired PS4.  Also ran QC sketches.  All was on the Minima board

### DIFF
--- a/avrpins.h
+++ b/avrpins.h
@@ -1320,6 +1320,135 @@ MAKE_PIN(P24, Pin_nRF51822_to_Arduino(D24));
 
 #undef MAKE_PIN
 
+#elif defined(__IMXRT1062__) && (defined(ARDUINO_TEENSY40) || defined(ARDUINO_TEENSY41))
+// Teensy 4.x
+
+#include "core_pins.h"
+
+#define MAKE_PIN(className, pin) \
+class className { \
+public: \
+  static void Set() { \
+    digitalWriteFast(pin, HIGH);\
+  } \
+  static void Clear() { \
+    digitalWriteFast(pin, LOW); \
+  } \
+  static void SetDirRead() { \
+    pinMode(pin, INPUT); \
+  } \
+  static void SetDirWrite() { \
+    pinMode(pin, OUTPUT); \
+  } \
+  static uint8_t IsSet() { \
+    return digitalReadFast(pin); \
+  } \
+};
+
+MAKE_PIN(P0, 0);
+MAKE_PIN(P1, 1);
+MAKE_PIN(P2, 2);
+MAKE_PIN(P3, 3);
+MAKE_PIN(P4, 4);
+MAKE_PIN(P5, 5);
+MAKE_PIN(P6, 6);
+MAKE_PIN(P7, 7);
+MAKE_PIN(P8, 8);
+MAKE_PIN(P9, 9);
+MAKE_PIN(P10, 10);
+MAKE_PIN(P11, 11);
+MAKE_PIN(P12, 12);
+MAKE_PIN(P13, 13);
+MAKE_PIN(P14, 14);
+MAKE_PIN(P15, 15);
+MAKE_PIN(P16, 16);
+MAKE_PIN(P17, 17);
+MAKE_PIN(P18, 18);
+MAKE_PIN(P19, 19);
+MAKE_PIN(P20, 20);
+MAKE_PIN(P21, 21);
+MAKE_PIN(P22, 22);
+MAKE_PIN(P23, 23);
+MAKE_PIN(P24, 24);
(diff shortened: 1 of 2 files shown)


xxxajk (changes requested):
Please reformat with tabs as 8 spaces, otherwise, looks great!
Optional/additional/bonus points consideration: make a new definition in settings.h that reduces the teensy4 elif, and places it in one single place, in order to reduce constant similar long if/elif blocks. This is something I would have done years ago, and still have not had a chance to do.

mjs513:
Sorry been sidetracked with other issues and projects.  Looking at both the Github version and my version - there are no tabs in the files only spaces.

Sorry.

xxxajk:
> Sorry been sidetracked with other issues and projects. Looking at both the Github version and my version - there are no tabs in the files only spaces.
> 
> Sorry.

Yes, but they don't appear to be 8 space indents...

Jtonna:
> Updated for latest Arduino R4 boards (maxima/wifi).
> 
> Tested BT with PS4 and XBox, and wired PS4. Also ran QC sketches. All was on the Minima board

Can confirm this allows compilation & uploads to go through, but in my 
attempts to run the USB_desc using these changes with the R4 Minima & an ARCELI USB Host Shield & am getting the "OSC Did Not Start" issue. I know the shield is working as I've tested it on 3 Leonardos.

May I ask which USB host shield are you using & was there any special config needed for OSC to start in your testing?

alejaxdro:
I just tested this with the Arduino R4 wifi and after getting the "OSC Did Not Start" issue, I remembered that this board requires power on VIN(I connected a 12v supply) when the Power Select Switch on the USB shield is set to VIN. Otherwise move the switch over to 5V and it also works fine, but I find this to pull too much current and the onboard regulator gets too hot. I'm also connecting to a hub and connecting 4 devices, so that would also push this setup over it's limit.

fraca7:
Hello. I was testing this on my new R4 Wifi board and got the "OSC did not start" error with both my shields (from different manufacturers). After digging a bit I found out the problem is at the SPI level; the board_qc sketch kept finding out the wrong die revision or failing at the "SPI long test" step, with the dreaded "0x55 pattern is transmitted via SPI" message.

To be honest I don't know anything about SPI but on a hunch I changed every occurence of

`        USB_SPI.beginTransaction(SPISettings(260000000, MSBFIRST, SPI_MODE0)); // The MAX3421E can handle up to 26MHz, use MSB First and SPI mode 0
`

in usbhost.h, replacing the 26000000 value with something else. It looks like when this value is <= 23500000 both of my shields work perfectly. Starting at 24000000 the board_qc sketch fails (so do the examples of course).

According to the Arduino forums this seems to happen to several people, so leaving this here for posterity.

xxxajk:
> Hello. I was testing this on my new R4 Wifi board and got the "OSC did not start" error with both my shields (from different manufacturers). After digging a bit I found out the problem is at the SPI level; the board_qc sketch kept finding out the wrong die revision or failing at the "SPI long test" step, with the dreaded "0x55 pattern is transmitted via SPI" message.
> 
> To be honest I don't know anything about SPI but on a hunch I changed every occurence of
> 
> ` USB_SPI.beginTransaction(SPISettings(260000000, MSBFIRST, SPI_MODE0)); // The MAX3421E can handle up to 26MHz, use MSB First and SPI mode 0`
> 
> in usbhost.h, replacing the 26000000 value with something else. It looks like when this value is <= 23500000 both of my shields work perfectly. Starting at 24000000 the board_qc sketch fails (so do the examples of course).
> 
> According to the Arduino forums this seems to happen to several people, so leaving this here for posterity.

There was another MCU released by microchip that had SPI speed limits because of poor silicon design.
That said--

UHS3.0  does a quick speed verification test, which not only avoids the entire issue, but also allows really lousy SPI lines (ones that are too long) to be used, within limits, of course...

The only bad thing about UHS3 is that it will never be able to run on the classic uno, and the mega1280/2560 just barely have enough room (RAM is too small!) USB is a heavy lift!

The only reason that UHS2 "works" at all is because it cheats -- it doesn't actually do the enumeration properly to spec.
In order to avoid high RAM usage these cheats can cause strange issues with certain combinations of device drivers. It's why some will never work, and why some combinations don't.

skhrlx:
Using this, my mouse can only get 125hz polling rate, when I use it on my leonardo, my mouse gets the 1000hz, do you know what I can do? The information about the r4 + usbhostshield is very limited for now

Lauszus (approved):

Lauszus:
Sorry for the long delay. Merging it now.

PR #818, opened 2024-07-14
Add support for 045e:0202 duke xbox controller and fix analog hat readings

herraa1:
The current code uses hardcoded endpoints to read/write to original Xbox controllers. This works for controllers like 045e:0289 which uses the same endpoint configuration as the hardcoded one.

045e_0289_device_descriptor
045e_0289_config_descriptor

But other original Xbox controllers like 045e:0202 do not follow the hardcoded configuration and end up not working (can't read reports from them).

045e_0202_device_descriptor
045e_0202_config_descriptor

This PR modifies the XBOXOLD code to determine endpoints by reading configuration descriptors as in the XBOXONE, instead of hardcoding them. This makes both kind of controllers work.
On this same PR a patch to properly read the left and right analog hats is included too.

### DIFF
--- a/Usb.cpp
+++ b/Usb.cpp
@@ -823,7 +823,12 @@ uint8_t USB::getConfDescr(uint8_t addr, uint8_t ep, uint8_t conf, USBReadParser 
 
         //USBTRACE2("\r\ntotal conf.size:", total);
 
-        return ( ctrlReq(addr, ep, bmREQ_GET_DESCR, USB_REQUEST_GET_DESCRIPTOR, conf, USB_DESCRIPTOR_CONFIGURATION, 0x0000, total, bufSize, buf, p));
+        /*
+	   At least 045e:0289 complains if nbytes is greater than total when calling ctrlReq().
+           Make sure that we don't request chunks greater than total length, now that XBOXOLD
+	   retrieves and parses configuration descriptors.
+	 */
+        return ( ctrlReq(addr, ep, bmREQ_GET_DESCR, USB_REQUEST_GET_DESCRIPTOR, conf, USB_DESCRIPTOR_CONFIGURATION, 0x0000, total, (total<bufSize)?total:bufSize, buf, p));
 }
 
 //get string descriptor
--- a/XBOXOLD.cpp
+++ b/XBOXOLD.cpp
@@ -47,6 +47,9 @@ const uint8_t XBOXOLD_BUTTONS[] PROGMEM = {
 XBOXOLD::XBOXOLD(USB *p) :
 pUsb(p), // pointer to USB class instance - mandatory
 bAddress(0), // device address - mandatory
+bNumEP(1), // If config descriptor needs to be parsed
+qNextPollTime(0), // Reset NextPollTime
+pollInterval(0),
 bPollEnable(false) { // don't start polling before dongle is connected
         for(uint8_t i = 0; i < XBOX_MAX_ENDPOINTS; i++) {
                 epInfo[i].epAddr = 0;
@@ -68,6 +71,7 @@ uint8_t XBOXOLD::Init(uint8_t parent, uint8_t port, bool lowspeed) {
         EpInfo *oldep_ptr = NULL;
         uint16_t PID;
         uint16_t VID;
+        uint8_t num_of_conf; // Number of configurations
 
         // get memory address of USB device address pool
         AddressPool &addrPool = pUsb->GetAddressPool();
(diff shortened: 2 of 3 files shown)


XBOXOLD.cpp
```diff
@@ -233,37 +250,108 @@ uint8_t XBOXOLD::Init(uint8_t parent, uint8_t port, bool lowspeed) {
...
 uint8_t XBOXOLD::Poll() {
+        uint8_t rcode = 0;
+
         if(!bPollEnable)
                 return 0;
-        uint16_t BUFFER_SIZE = EP_MAXPKTSIZE;
-        pUsb->inTransfer(bAddress, epInfo[ XBOX_INPUT_PIPE ].epAddr, &BUFFER_SIZE, readBuf); // input on endpoint 1
-        readReport();
-#ifdef PRINTREPORT
-        printReport(BUFFER_SIZE); // Uncomment "#define PRINTREPORT" to print the report send by the Xbox controller
```
Lauszus, line 252: The code below is now basically the same as `printReport`, so we can just remove it and call `printReport`

XBOXOLD.cpp
```diff
@@ -233,37 +250,108 @@ uint8_t XBOXOLD::Init(uint8_t parent, uint8_t port, bool lowspeed) {
...
                 buttonValues[i] = readBuf[i + 4]; // A, B, X, Y, BLACK, WHITE, L1, and R1
 
-        hatValue[LeftHatX] = (int16_t)(((uint16_t)readBuf[12] << 8) | readBuf[13]);
-        hatValue[LeftHatY] = (int16_t)(((uint16_t)readBuf[14] << 8) | readBuf[15]);
-        hatValue[RightHatX] = (int16_t)(((uint16_t)readBuf[16] << 8) | readBuf[17]);
-        hatValue[RightHatY] = (int16_t)(((uint16_t)readBuf[18] << 8) | readBuf[19]);
+        hatValue[LeftHatX] = *(int16_t *)(p + 12);
+        hatValue[LeftHatY] = *(int16_t *)(p + 14);
+        hatValue[RightHatX] = *(int16_t *)(p + 16);
+        hatValue[RightHatY] = *(int16_t *)(p + 18);
```
Lauszus, line 354: I guess the bug here was the data is little-endian.

Please use shifts, as before, so this also works on big-endian MCUs.

Usb.cpp
```diff
@@ -823,7 +823,12 @@ uint8_t USB::getConfDescr(uint8_t addr, uint8_t ep, uint8_t conf, USBReadParser
 
         //USBTRACE2("\r\ntotal conf.size:", total);
 
-        return ( ctrlReq(addr, ep, bmREQ_GET_DESCR, USB_REQUEST_GET_DESCRIPTOR, conf, USB_DESCRIPTOR_CONFIGURATION, 0x0000, total, bufSize, buf, p));
+        /*
+	   At least 045e:0289 complains if nbytes is greater than total when calling ctrlReq().
+           Make sure that we don't request chunks greater than total length, now that XBOXOLD
+	   retrieves and parses configuration descriptors.
+	 */
```
Lauszus, line 830: Won't it be better to increase `bufSize`?

This code would get an incomplete configuration descriptor if `total` is greater than `bufSize`.

XBOXOLD.cpp
Lauszus: Also this does not compile: https://github.com/felis/USB_Host_Shield_2.0/actions/runs/9931283871/job/27451495122?pr=818#step:7:315

Usb.cpp
herraa1: Maybe my comment is not clear, I'll try to explain what's the issue.

The problem is that the device errors out if you try to read more bytes than the total configuration descriptor total length.

The normal way of reading such a descriptor is you first read the configuration descriptor with a small enough length (minimum 9 bytes is required because some controllers complain if you read less than the basic configuration descriptor information size) to retrieve the actual total length `wTotalLength` of the configuration descriptor. Then, once you know the total length, you read again the full configuration descriptor using the total length.

But, and here's the thing, if you request more data than the total configuration descriptor length, the device complains (and the current code tries to read 64 bytes when the configuration descriptor is just 12 bytes).

So the fix is to not read more than the total configuration descriptor length (and of course we need a buffer properly sized).

The readBuf size is 64 bytes, the total configuration descriptor length is 12 bytes in this case, so we need to use the readBuf but limit the total data requested to 12 bytes for the control request to work.

Increasing the buffer size would help only in case we know that we need to read a larger than 64 bytes configuration descriptor.
With the current proposed fix it is true that we could get an incomplete configuration descriptor if `total` is greater than `bufSize`, as we won't be able to read the full descriptor in the buffer. But that is a consequence of the finite buffe size (and the code prevents overflowing the buffer if the total length is greater than the buffer size).

If the current buffer size has been ok until now, then I think it is fine to keep its size as it is today. We don't need a larger buffer for the devices involved.

XBOXOLD.cpp
herraa1: Yes, you are right. I'll fix it.

XBOXOLD.cpp
herraa1: Yes, it looks like I was too accustomed to gcc and others enabling by default void pointer arithmetic.
I'll rework it using shifts and ORs. Thanks.

herraa1:
I've submitted two additional commits to the PR to fix the indicated concerns (printReport open coding and analog hat endianness fix using void pointer arithmetic).

herraa1:
Do you want me to squash the fixes for raised concerts into the feature commits to keep history simple and avoid non-working points in repository history? (I'm not so used to github and established workflows, although I have raw git experience).

herraa1:
I ended up squashing the fixes into their original commits (rebase).
Please, let me know if you need something else before accepting the PR. Thanks.

XBOXOLD.cpp
herraa1: Fixed.

XBOXOLD.cpp
herraa1: Fixed.

Lauszus (approved):
Thanks!

Lauszus:
> I ended up squashing the fixes into their original commits (rebase). Please, let me know if you need something else before accepting the PR. Thanks.

Looks great. Thanks for cleaning up the commits :)

Lauszus:
Btw don't worry about the CI. I'll fix that.

PR #857, opened 2026-05-19
Add support for Logic Green LGT8F328p

Hummtaro:
Added support for Logic Green LGT8F328p.

The LGT8F328p is largely a clone of the Atmega328p but with some nice enhancements.
It's specified to operate at up to 32MHz at 3.3V and is available in Nano, Uno and Pro Mini style board formats, making it ideal for the USB Host Shield. (Why does it seem entirely unknown here?)

It's an AVR and the pinout is identical to the classic Arduino's. Therefore i only had to touch the multi-byte write function to use SPI.transfer() instead of writing SPDR directly, since the SPI buffer works slightly different. (SPDR is a 4-byte register and it got an additional status register.)


Lauszus (approved):

Lauszus:
Thanks!

PR #860, opened 2026-08-20
Adds support for AVR64DD32

flowernert:
Adds support for AVR64DD32 and lays the ground to implement for other AVR Dx MCUs.
Compiled and functionally tested with qc_board sketch on AVR64DD32, official Arduino Uno (328p) and Arduino micro (32U4) clone.


avrpins.h
```diff
@@ -41,6 +41,15 @@ e-mail   :  support@circuitsathome.com
 #define BOARD_BLACK_WIDDOW
 #endif
 
+// Port definition configuration for classic AVR MCUs or newer Dx AVR MCUs
+#if !defined(DXCORE)
+  #define PORT_AVR_CLASSIC
+#elif defined(DXCORE)
+  #define PORT_AVR_DX
+#else
+  #error "unknown AVR port definition, you need to add support for the target MCU"
```
Lauszus: This will never get hit. Just replace this with:

```C
#ifdef DXCORE
#define PORT_AVR_DX
#else
#define PORT_AVR_CLASSIC
#endif
```

avrpins.h
```diff
@@ -114,6 +122,18 @@ e-mail   :  support@circuitsathome.com
       enum{Width=sizeof(DataT)*8};\
     };
 
+//Port definitions for AVR Dx family
+#define MAKE_PORT_AVR_DX(portName, className, ID) MAKE_PORT_AVR_CLASSIC(portName.OUT, portName.DIR, portName.IN, className, ID)
+
+// Indirection layer to add support for AVR ports software architecture variants
+#if  defined(PORT_AVR_CLASSIC)
```
Lauszus:
```suggestion
#if defined(PORT_AVR_CLASSIC)
```

avrpins.h
```diff
@@ -114,6 +122,18 @@ e-mail   :  support@circuitsathome.com
...
+  #define MAKE_PORT(portName, ddrName, pinName, className, ID) MAKE_PORT_AVR_CLASSIC(portName, ddrName, pinName, className, ID)
+#elif defined(PORT_AVR_DX)
+  #define MAKE_PORT(portName, ddrName, pinName, className, ID) MAKE_PORT_AVR_DX(portName, className, ID)
+#else
+  #error "avr core or platform not defined, you'll probably have to add some code to support it"
```
Lauszus: Please write a better error message

Lauszus (commented):
Thanks! Just left a few comments :)

avrpins.h
flowernert: yeah I know that it won't, however it was to lay a safe path if more port definition variants continue to come in the future. But I get that it might be unnecessary now so fine to remove it

flowernert:
Thanks for your feedback, I corrected and pushed the changes based on your suggestions.

